PR #11, opened 2015-03-27
Add support for hybrid client

fbodz:
Not sure if this is the way you wanted to accomodate the hybrid client @feross but it appears to work with browserify and on the desktop. Node-webrtc is adding preliminary support for precompiled binaries so hopefully this will make it easier to install. I am interested in your thoughts surrounding the hybrid client. Thanks.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,5 @@
 module.exports = Peer
+module.exports.hybrid = false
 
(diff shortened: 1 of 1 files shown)


feross:
Thanks for this PR. I added node.js support in a different way. This is released as 4.2.0. Best!

PR #12, opened 2015-03-30
add options for offer and answer constraints

maxogden:
needed for a scenario where initiator is e.g. audio only, but other side responds with audio + video streams


feross:
Thanks! Released as 4.1.0.

PR #15, opened 2015-04-17
wait a bit before destroying on end

mafintosh:
this is a bit of a hack but seems to work in most cases. basically there is a bug right right now where simple-peer will discard data when you call `.end()` this happens because we immediately call `.close()` on the datachannel in the flush event and this seems to forcible destroy the datachannel before it has finished flushing

in the future we should look into better ways of doing this :/


feross:
Released as 4.2.3.

PR #16, opened 2015-04-19
opts.wrtc

### LINKED ISSUE #14: optionalDependency on wrtc is unwanted for web projects
Hi @feross 

Continuing the discussion in https://github.com/js-platform/node-webrtc/issues/194 (and the clutter of trying to solve it with --no-optional) - I see another downside of defining wrtc as an optionalDependency of simple-peer.

I have several projects that depends on simple-peer, because it's awesome :) These projects only use it in web pages, but have no interest for the webrtc stack for nodejs, and they get installed on cloud platforms only to serve those pages to users.

Now wrtc is always attempted to install along with simple-peer, and since wrtc is not yet providing binaries for all platforms, this also starts to build from source... and fails on occasion. So I think wrtc is still not yet in a stable status, but even if it was it is definitely not a small nifty library that can be dragged along anyplace with simple-peer (which is a small and nifty library).

Maybe use [peerDependency](https://docs.npmjs.com/all#peerdependencies) instead so that if anyone wants to use simple-peer with wrtc it will simple depend on both?

Thanks!

substack:
This pull request fixes #14 by removing the hardly-optional-dependency on wrtc, which creates a lot of build and installation headaches for whipping up a simple webrtc app that doesn't even use node. Out of 4 people trying to build a webrtc app with simple-peer last thursday at sudoroom, nobody could get simple-peer to successfully install.

Instead, with this patch you can pass `opts.wrtc` as a parameter in node. In the browser you will get the detected version like before but you can also pass in a custom `opts.wrtc` implementation in the browser too.

The interface changes slightly, so this is a breaking change, but the changes are relatively minor and only apply in node:

```js
var SimplePeer = require('simple-peer')
var wrtc = require('wrtc')

var peer1 = new SimplePeer({ initiator: true, wrtc: wrtc })
var peer2 = new SimplePeer({ wrtc: wrtc })
```

I think the slight extra penalty in API is more than made up for with the gains in installation issues. Plus, by passing in wrtc as an option, the version of wrtc can be versioned separately from simple-peer. This means bugs can be fixed upstream without updating simple-peer and simple-peer can be upgraded while still using an older version of wrtc that is known to work on a platform. And once there is a lighter-weight version of wrtc available, it will be easy to swap it in via `opts.wrtc`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,17 @@ function gotMedia (stream) {
 }
 ```
 
-For two-way video, simply pass a `stream` option into both `Peer` constructors. Simple!
+### in node
+
+To use this library in node, pass in `opts.wrtc` as a parameter:
+
+```js
+var SimplePeer = require('simple-peer')
+var wrtc = require('wrtc')
+
+var peer1 = new SimplePeer({ initiator: true, wrtc: wrtc })
+var peer2 = new SimplePeer({ wrtc: wrtc })
+```
 
 ## production apps that use `simple-peer`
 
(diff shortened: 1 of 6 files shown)


feross:
Thanks - this is a clean solution to the installation issues. Too bad `wrtc` doesn't build more reliably...

feross:
Released as 5.0.0.

PR #22, opened 2015-05-10
Allow options to be passed into PeerConnection.createDataChannel

kevzettler:
So you can create unreliable data channels like: `{ordered: false, maxRetransmits: 0 }`

http://html5index.org/WebRTC%20-%20RTCDataChannelInit.html
http://w3c.github.io/webrtc-pc/#dictionary-rtcdatachannelinit-members

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,6 +29,7 @@ function Peer (opts) {
     config: Peer.config,
     constraints: Peer.constraints,
     channelName: (opts && opts.initiator) ? hat(160) : null,
+    dataChannelOpts: {},
     trickle: true,
     allowHalfOpen: false, // duplex stream option
     highWaterMark: 1024 * 1024 // duplex stream option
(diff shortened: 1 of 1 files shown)


maxogden:
@kevzettler oh nice, i was wondering how to do this. I remember a couple of years ago before they had SCTP in the stack you only had unreliable. Now that SCTP is there it seems like you only have reliable. The two options you have in your example here make sense, but do they fully emulate UDP-esque unreliability? I wonder if you can force UDP (by e.g. removing TCP candidates from the SDP) and disable SCTP reliability altogether (maybe thats what these two options do but it seems like there would be more than just these two).

P.s. I am gonna defer to @feross' judgment on this merge (LGTM), I am just asking this stuff out of curiosity

feross:
LGTM :+1: Merged and released as 5.5.0, but I renamed the option to `channelConfig` for parity with the existing `channelName` option, and updated the readme.

Max, let's cc @fippo and @DamonOehlman who know more than me and might have answers to your questions.

PR #23, opened 2015-05-10
fixed documentation error for Peer.send

jraoult:
`Peer.send` doesn't actually serialise the given argument with `JSON.stringify`


feross:
Thanks for reporting this issue. This is actually a regression introduced in https://github.com/feross/simple-peer/commit/bf697c17d6d12a83aa71e698c42a3f03b9fead23, not a documentation error.

Fixed and released as 5.5.0.

jraoult:
@feross cool, I was wondering either the doc is stale either the impl si broken.

Issue #20, opened 2015-05-03
Offer a way to munge the "local" offer


jraoult:
Would be great to give a way to patch the offer's SDP before setting it as a local description (like it is done for the "speed hack"). May be a function passed as an option to the constructor like:
```javascript
new SimplePeer({
  sdpMunger: function(sdp) {
    // transform the SDP
    return sdp;
  }
})
```

I would need it to change to configuration of the audio codec (set it in stereo mode and change the `ptime`).

feross:
Yep, I see the use case for this. Let's name the property `sdpTransform` though. PR welcome!

jraoult:
@feross got you ;-) I'll see what I can do.

contra:
Can you link to the "speed hack" for reference?

feross:
This is the "speed hack": https://github.com/feross/simple-peer/blob/4bec667b979e5f6baa2e15d3ba09d82890047584/index.js#L490-L493

feross:
This is supported as as of 5.8.0.

PR #25, opened 2015-05-18
Adding sdp transform

jraoult:
#### What's this PR do?
It allows the API user to transform the SDP before being applied locally for both offer and answer

#### What is the relevant tickets?
see #20 

### How to use it?
Here a example of what can be done:
```javascript
[...]
transformSdp: (sdp) {
    var sdpObj = sdpTransform.parse(sdp);
    // turn on stereo
    sdpObj.media[0].fmtp[0].config = 'stereo=1;';
    return sdpTransform.write(sdpObj);
  }
};
[...]
```


jraoult:
@feross sorry for the delay but here it is :-)

feross:
Thanks for the PR! Released as 5.8.0.

PR #28, opened 2015-05-27
pick the right attributes off of each signal object (closes #26)

### LINKED ISSUE #26: shim support
There are some shims (for example: https://code.google.com/p/webrtc4all/) that allow you to do webrtc all the way back to IE6. Crazy.

The API is *supposed to* be the same as the spec but it looks a little different from a first glance. Not going to be easy to drop it right in. I was curious if you had any ideas for how this should work with simple-peer

Stuff I can think of off the top of my head:

- Create a layer that makes this shim conform to the spec, write to window and simple-peer will use it unknowingly as long as it behaves the same
- simple-peer takes an implementation of webrtc as an option, but defaults to the browser's specified version
- simple-peer adds support for this shim outright

LMK your thoughts :+1:

contra:
some shims provide objects that can't be serialized by JSON.stringify (temasys)


feross:
Released as 5.10.0.

PR #31, opened 2015-07-10
Cache received ice candidates if setting remote description has not c…

areiter:
As the addIceCandidate call completes asynchronously it happens from time to time that ICE candidates are received from the remote side, but the remote description is not set (yet!).

Observed in NodeJS with wrtc. I have not tested if the problem also occurs in the browser.

### DIFF
--- a/index.js
+++ b/index.js
@@ -64,6 +64,7 @@ function Peer (opts) {
   self._channelReady = false
   self._iceComplete = false // ice candidate trickle done (got null candidate)
   self._channel = null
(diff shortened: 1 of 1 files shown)


feross:
Released as 5.11.4. Thanks!

feross:
It's now in the standard that addIceCandidate should be queued along with the rest of the methods, so this fix won't be necessary once browsers update to match the spec.

https://webrtcstandards.info/addicecandidate-queued-method/

PR #32, opened 2015-07-15
Added socket.io-p2p to list of production apps

ArtskydJ:
Added `socket.io-p2p` to list of production apps using `simple-peer`.


maraoz:
nice!

feross:
This is great!

PR #54, opened 2015-12-11
remove onnegotionneeded workaround, firefox now emits this

contra:
https://bugzilla.mozilla.org/show_bug.cgi?id=1017888

Need feedback if this also applied to any other implementations (wrtc?)


contra:
Tested on a few implementations, this is still needed for: temasys, wrtc

feross:
Can we still make a change so that this hack is not used in Firefox?

PR #61, opened 2016-01-28
Fix bug with wrtc not handling Buffers correctly

paulkernfeld:
In this pull request, I also added back in the `wrtc: wrtc` arguments in the tests, which I see that you recently removed. If you want those to stay out, I can remove that commit from this PR.

### DIFF
--- a/index.js
+++ b/index.js
@@ -197,7 +197,7 @@ Peer.prototype.send = function (chunk) {
   }
 
   // `wrtc` module doesn't accept node.js buffer
-  if (Buffer.isBuffer(chunk) && !isTypedArray.strict(chunk)) {
+  if (Buffer.isBuffer(chunk)) {
     chunk = new Uint8Array(chunk)
   }
 
--- a/index.js
+++ b/index.js
@@ -115,6 +115,14 @@ function Peer (opts) {
 
 Peer.WEBRTC_SUPPORT = !!getBrowserRTC()
 
+try {
+  require('wrtc')
+  Peer.USING_WRTC = true
+} catch (err) {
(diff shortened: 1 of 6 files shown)


paulkernfeld:
I tested this locally against Chrome 47 and Firefox 43 on OSX, and it passed.

feross:
The commits to add zeroes the binary test and make the tests easier to run on node are fine.

But the core of the PR, I'm not sure about. This will cause an extra copy on every send, even in non-wrtc situations, which is likely to have performance implications.

paulkernfeld:
Oh, I see. Yeah, that seems like it could be bad. Would you be opposed to code that special-cases wrtc and only converts from a Buffer in that case? Or, is there some way to convert data from a Buffer without doing a copy?

Another solution might be a simple polyfill library that tries to abstract away some of this stuff.

feross:
I think a bit of code that special cases `wrtc` is the best way to go.

There's no real way to solve this for `wrtc` without a copy it seems, since the problem is that it doesn't like the subclassed `Uint8Array`.

paulkernfeld:
All right, here's a commit that special-cases wrtc. I think there are probably a million ways to do this, so I'm happy to change this if there's a better way.

feross:
Thanks for updating your PR. I think I found a cleaner way to do this for now. Let's just use the existence of the `wrtc` option to detect wrtc. It's currently the only implementation in node anyway.

Once another viable alternative exists in node, or the issue is fixed in `wrtc`, we can revisit this.

Published as 5.11.9.

paulkernfeld:
Awesome!

PR #71, opened 2016-03-18
Fix destroy error

Autarc:
Currently the public `.destroy` signature doesn't match with the internal `._destroy`. As other libraries like [bittorrent-swarm](https://github.com/feross/bittorrent-swarm/blob/master/index.js#L133) are using the CSP interface and pass an error as the first argument, descendants of EventEmitter complain about the invalid parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,11 +240,11 @@ etc.), `ArrayBuffer`, or `Blob` (in browsers that support it).
 Note: If this method is called before the `peer.on('connect')` event has fired, then data
 will be buffered.
 
-### `peer.destroy([onclose])`
+### `peer.destroy([err], [onclose])`
 
 Destroy and cleanup this peer connection.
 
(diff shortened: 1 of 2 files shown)


feross:
> CSP interface

What is that?

I think we don't want to expose an `err` parameter like this. It's not really useful the user to be able to destroy with an error like this, except for internal reasons. That's why there's a public `destroy()` and a private `_destroy()`.

I just fixed bittorrent-swarm to not do this anymore: https://github.com/feross/bittorrent-swarm/commit/d3640c6bb91ce661af11b0e2c7142935abf1f4b7

If there are other places you notice this, please open new issues! Thanks for the good work!

Autarc:
Sorry for the typo, meant a continuous passing style like its used for async operations and often seen with an error as the first parameter to propagate it.

Keeping it internal seems reasonable and will be the way to go from now on.

feross:
Cheers, thanks

Issue #68, opened 2016-03-07
Create renegotiation method


cusspvz:
This should allow us to:
- expose method to add streams `.addStream` #50
- allow mutes and video interruption on the source side

@feross has any work been initiated regarding renegotiation?

feross:
How do you propose this would work? Can you provide a complete code example and explain a bit more.

cusspvz:
@feross sorry for my late response, had to switch priorities. I've written a renegotiation capable software two years ago, but I cannot recall how it worked, have to dig in.
I've already read your codebase, and I have **two ideas** for an implementation but I have to test the first one, because I don't know if it would work:

* The first one is trying to re-use the same `PeerConnection`, by handling streams on it and creating offers and answers as soon as we spot a renegotiation is needed, which I think is impossible but is worth to try, and if we have success, it would be the easiest implementation.

* The second one is to abstract `PeerConnection` interface from your current API, this way we could refactor some code and fix some bugs you had (related with stream destroying) without causing Breaking Changes on your current API design. This way there would be two main classes:
  * `SimplePeer`
    * Streamable Passthrough from the `currentPeerConnection`;
    * PeerConnections Manager (handling renegotiations against streaming changes and  long-reconnecting delays);
    * Signalling transporter gate from all `PeerConnections` (so we can initiate new ones before having data flowing through)
  * `PeerConnection` - Streamable;

cusspvz:
On the second approach, I think it could be a good Idea to have separated PeerConnections for different streams, it would make renegotiation easier.

feross:
@cusspvz Thanks for explaining your thoughts. This all sounds really complicated. If there's no way to do this simply using the same RTCPeerConnection, then I don't think this belongs in `simple-peer`. We can try to expose the right events so that this can be built as a layer on top of `simple-peer`.

cusspvz:
Yeah, I admit the second one seems a little tricky to implement, besides, its complexity doesn't go well along with this repo's purpose...

I think I can write out some tests and start playing along with the first approach until the end of this week.

feross:
@cusspvz Sounds good. Happy to check out what you come up with!

cusspvz:
Already PR'ed some changes! :)
Seems the first approach worked, the only issue I'm having is related with firefox and I think it is because of testing with fake MediaStreams, but I have to dig a little bit more to have sure.

Unfortunately I'm running out of time to work on it today, I can only finish this up on Monday.

[...]

Issue #50, opened 2015-12-05
expose addStream on prototype


contra:
right now you can only specify the stream in the constructor

you should be able to do it lazily after the connection has been opened, that way you can start the connection process as soon as possible instead of handling camera access and connection in series

feross:
The downside of creating streams lazily is then you have to handle renegotiation.

RationalCoding:
As an alternative to renegotiation, it's possible to merge MediaStream before passing them into simple-peer. You can add/remove streams to this merged stream as you please, without renegotiation.

Not the ideal solution, but it's a workaround until the spec around multistreaming stablizes. 

[Demo](https://rationalcoding.github.io/video-stream-merger/p2p.html)

ben-pr-p:
To force renegotation, I've been destroying and recreating Peer objects, and although this may involve slightly more renogiation than necessary it's been working for me.

If you're using react, check out the component I've wrapped it in to handle that logic for me https://github.com/ben-pr-p/assemble/blob/master/client/room/user-blob/webrtc/react-simple-peer.js

If you pass the component a new stream, it will handle renegotiation by destroying and recreating a Peer object internally.

tvedtorama:
@RationalCoding I think your library looks like a promising solution for this problem, but I have a few questions:

* Is the merging logic supported on the browsers that support WebRTC or just on Chrome?
* Could you publish the source for your demo, I can't find it?
* Is it possible to detect that there are "no streams in the merged stream"?  Ie - that the host is currently not transmitting anything?
  * I guess this could be sent alongside the stream, on the data channel, but anyways...

cusspvz:
Being solved at #250

RationalCoding:
Implemented in #250.

PR #72, opened 2016-03-19
[WIP] Implementation of .renegotiate and .negotiate API

cusspvz:
- [x] Add `.renegotiate`
- [x] Add `.negotiate`
- [x] Add `negotiate` event
- [x] Add `.addStream`
- [ ] Add `.removeStream`
- [ ] Add `.addDataStream` - have a question if we should allow multiple datastreams
- [ ] Add `.removeDataStream`
- [x] Add tests where a fake MediaStream instance is added into peers post their first established connection
  - [x] Added `.editorconfig`
  - [x] Added `node_modules/` to `.gitignore` and `.npmignore`

Current stage:
- firefox is giving an error because of differences on sdp both m lines, unfortunately I can only finish this implementation Monday

Related issues:
- #68
- #50

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+# http://editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+max_line_length = 80
+trim_trailing_whitespace = true
+
+[*.md]
+max_line_length = 0
+trim_trailing_whitespace = false
+
+[COMMIT_EDITMSG]
+max_line_length = 0
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
-img/
 node_modules/
--- a/.npmignore
+++ b/.npmignore
@@ -1,2 +1,4 @@
 img/
 node_modules/
+.*
+*.log
--- a/test/renegotiation.js
+++ b/test/renegotiation.js
@@ -1,0 +1,86 @@
+var common = require('./common')
+var Peer = require('../')
+var test = require('tape')
+
+var getUserMedia = typeof navigator !== 'undefined' && (
+  navigator.webkitGetUserMedia && function ( options, fulfill, reject ) {
+    // Create a fake stream
+    var stream = new webkitMediaStream()
+    Object.defineProperty(stream, 'active', { value: true })
+
+    fulfill( stream )
+  } ||
+  navigator.mozGetUserMedia && function ( options, fulfill, reject ) {
+    options.fake = true
+    navigator.mozGetUserMedia( options, fulfill, reject )
+  } ||
+  navigator.getUserMedia
+)
+
+var mediaStream
+test('fetch fake media stream', function (t) {
+  getUserMedia({ video: true, audio: true },
+    function ( stream ) {
+      mediaStream = stream
+      t.end()
+    },
+    function ( err ) {
+      t.fail( 'error fetching mediaStream' )
+      t.end()
+    }
+  )
+})
+
+var config
+test('get config', function (t) {
+  common.getConfig(function (err, _config) {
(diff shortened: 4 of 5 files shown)


cusspvz:
NOTE: I've seen that there are linting issues on the CD, I will fix them just before landing the PR.

.editorconfig
```diff
@@ -0,0 +1,17 @@
...
+indent_style = space
+max_line_length = 80
+trim_trailing_whitespace = true
+
+[*.md]
+max_line_length = 0
+trim_trailing_whitespace = false
+
+[COMMIT_EDITMSG]
+max_line_length = 0
```
feross: Please don't commit this file

.gitignore
```diff
@@ -0,0 +1 @@
+node_modules/
```
feross: This is not necessary. Add this to your own global gitignore

.npmignore
```diff
@@ -1 +1,4 @@
 img/
+node_modules/
```
feross: npm always ignores node_modules, not necessary

.npmignore
```diff
@@ -1 +1,4 @@
...
+.*
+*.log
```
feross: What generates these files? Remove this please

feross:
I think it's a mistake to stop using the onnegotiationneeded event. That was required for things to work for me before -- has that changed?

cusspvz:
I have to finish this up, the only issue I had was with the Firefox implementation, tests were flawless on chrome.
Regarding your question related with `onnegotiationneeded`, it is called whenever a `MediaStream` is added, perhaps you had async issues and you were calling offer before having the stream?

Seen that you've closed the PR, are you still interested in having renegotiation here?

feross:
Sorry for closing the PR -- that was an accident. Before I can merge this, you need to fix the merge conflicts, fix the code style, and fix it on Firefox.

Looks like you also might need to add:

```
if (renegotiating) return
```

to the top of the `_onIceCandidate` function for Chrome.

See: https://www.webrtc-experiment.com/docs/how-to-switch-streams.html

feross:
Also, the test style is kind of weird. Can you clean it up a bit?

test/renegotiation.js
```diff
@@ -0,0 +1,86 @@
+var common = require('./common')
+var Peer = require('../')
+var test = require('tape')
+
+var getUserMedia = typeof navigator !== 'undefined' && (
+  navigator.webkitGetUserMedia && function ( options, fulfill, reject ) {
```
feross: Just use an if statement. That would be much simpler.

MatthieuLemoine:
@cusspvz Are you still working on it ?

princemaple:
Hi @cusspvz , are you not interested in this anymore?

cusspvz:
@MatthieuLemoine @princemaple I am but not for now, but can arrange some time to inject some changes on the PR.
As far as I can remember, there was an issue on Firefox upon a renegotiation. 

Also, seems firefox only accepts one data stream per RTConnection.

@feross if I implement `addDataStream` should it error whenever it isn't possible?

jeregrine:
addStream is going to be deprecated in future versions. Firefox supports addTrack. They have a migration plan with chrome back-support here https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addStream

RationalCoding:
I think we should leave addStream in the simple-peer API. It's much simpler than adding individual tracks (although a secondary track API couldn't hurt). We already use addTrack internally.

jeregrine:
addStream is being deprecated. If simple-peer wants to support an addStream that simply calls addTracks on each track then that would be fine. But long term the API is going away.

cusspvz:
Closing due to #250 
Thanks for putting your efforts on it @RationalCoding.

PR #83, opened 2016-07-17
Added signal callback with answer

grkblood13:
Added callback to signal function that returns answer from offer.

### DIFF
--- a/index.js
+++ b/index.js
@@ -157,7 +157,7 @@ Peer.prototype.address = function () {
   return { port: self.localPort, family: 'IPv4', address: self.localAddress }
 }
 
-Peer.prototype.signal = function (data) {
+Peer.prototype.signal = function (data,callback) {
   var self = this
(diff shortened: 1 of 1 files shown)


feross:
@grkblood13 What is wrong with just listening for the `'signal'` event? I don't think we need multiple ways to get the signaling data in the API. This is just extra complexity.

Can you clarify the purpose of this PR?

grkblood13:
In my scenario I'm running one common peer that's server-side along with a socket.io signalling server and a hosted page that creates a browser peer for each connection. Since the signalling server's individual socket instances are out of of scope of the common 'signal' listener there's not a simple way to pair the answers generated from the offer function with the server-side socket instances they were initiated from. With this callback the answer from the offer is available in scope of the socket.

feross:
@grkblood13 I don't understand why if you're able to call `peer.signal(data, cb)` you can't call `peer.signal(data); peer.on('signal', cb)` In both cases, all that's needed is for `peer` to be in scope.

Issue #84, opened 2016-08-08
onaddstream is deprecated! Use peerConnection.ontrack instead.


KaeKun:
I got that warning on firefox console and I just replaced onaddstram by ontrack directly on the simplepeer.min.js and it didn't come out again. I think that'd be all to do?

feross:
Confirmed. PR sent: #85

PR #85, opened 2016-08-08
Fix "onaddstream is deprecated" warning

feross:
Fixes https://github.com/feross/simple-peer/issues/84


feross:
Would have been nice to get a review on this, but will merge now 👍

Issue #116, opened 2017-01-09
Support react-native-webrtc getStats


bkonkle:
In the `getStats` call, here:

https://github.com/feross/simple-peer/blob/master/index.js#L413-L438

In React Native using `react-native-webrtc`, the logic falls back to the Chrome implementation with the first two arguments inverted (#lolchrome) rather than following the standard method signature. `react-native-webrtc` is a good citizen and implemented it the [right way](https://github.com/oney/react-native-webrtc/blob/master/RTCPeerConnection.js#L158), but since the logic falls back to the Chrome block it doesn't work (and consequently 'connect' is never emitted).

Since Chrome is the known non-standard API here, I think a better approach would be to look for a Chrome-specific marker and then fall back to the standard signature that Mozilla uses.

I'll prepare a pull request, but I'd love any initial feedback you have in the meantime. 😄 

Thanks!

bkonkle:
caniuse.com says Chrome still uses the `webkit` prefix, so I'll do a quick fork inverting the logic and looking for webkitRTCPeerConnection.

feross:
I did what you suggested and we fallback to the standard behavior now. However, to fix a warning in Firefox, I changed that same standard behavior to use the promise-based version of `getStats`, since the callback version has been deprecated. So I'm guessing this won't work for you now.

feross:
Okay, third times the charm. Published a version that falls back to the callback version for `react-native-webrtc`.

6.1.3.

PR #117, opened 2017-01-09
Detect webkitRTCPeerConnection and fallback to standard

bkonkle:
See #116 for context. This detects `webkitRTCPeerConnection` to determine if Chrome's non-standard `getStats` signature needs to be used, and falls back to the standard signature that Mozilla and `react-native-webrtc` use.


bkonkle:
Tested in React Native iOS v0.39.2, and the latest Electron. I don't have a web project to test it with at the moment.

feross:
I was actually just editing this code today to fix a warning in Firefox (#118). Your PR makes a good point though, we should only special-case Chrome and fallback to the standard behavior. I'll update that.

feross:
@bkonkle Can you test 6.1.2 and see if your issue is resolved?

bkonkle:
6.1.3 Works great for my use case. Thanks!

feross:
Great!

Issue #122, opened 2017-01-17
Let's solve the recent getStats() instability


feross:
I've been playing with the code that decides which version of `RTCPeerConnection.prototype.getStats` gets called to add support for , and it's unfortunately pretty finicky.

There are at least 3 implementations that I know of:

- Single-parameter callback-based getStats() (non-standard)
  - Used by Chrome, `electron-webrtc`, `node-wrtc`
- Promise-based getStats() (standard)
  - Used by Firefox, Chrome (if "Experimental Web Platform Features" is enabled)
-  Two-parameter callback-based getStats() (deprecated, former standard)
  - Used by `react-native-webrtc`

It's not easy to feature detect which version is available, so we have to resort to user agent sniffing.

I propose that we:

- [x] Add an explicit `peer._browserDetect()` function so all the browser sniffing nastiness is isolated to a single function.
- [x] Re-enable the node.js tests that use `node-wrtc`, to catch regressions. (Hopefully it's easier to build these days).
- [x] Add node.js tests for `electron-webrtc`.
- [ ] Consider adding a `try-catch` around the `getStats` calls in case detection fails, or a user is using a very old or new version of a browser that has switched implementations (or a Chrome user who turned on "Experimental Web Platform Features")
- [ ] Encourage package maintainers to switch to the standard promise-based APIs by opening issues. Callback methods are non-standard/deprecated.

feross:
First PR, review please: https://github.com/feross/simple-peer/pull/123

SourceBoy:
I vote for `peer._browserDetect()` as the API differences between browsers have almost always been addressed this way in the past via a normalization layer by various vendors.

feross:
Okay, this is all fixed now! Lots of good stuff is in here now, including support for the new spec-compliant `getStats()` report format.

This new format is available in Chrome today if you enable "Experimental Web Platform Features". It will ship in Chrome 58 and `simple-peer` will be ready for it. I'm not sure when Firefox is shipping a spec-compliant format, but this should work for that too.

feross:
6.2.0

PR #123, opened 2017-01-17
Improve getStats() stability

feross:
For: https://github.com/feross/simple-peer/issues/122

- [x] Add an explicit `peer._browserDetect()` function so all the browser sniffing nastiness is isolated to a single function.
- [x] Re-enable the node.js tests that use `node-wrtc`, to catch regressions. (Hopefully it's easier to build these days).
- [x] Add node.js tests for `electron-webrtc`.


index.js
```diff
@@ -447,6 +447,24 @@ Peer.prototype.getStats = function (cb) {
       })
       cb(items)
     }, function (err) { self._onError(err) })
+}
+
+// Detect the WebRTC implementation. We only need this in a few places where
+// feature-detection does not work correctly.
+Peer.prototype._browserDetect = function () {
```
SourceBoy: Using a `switch (true)` statement for the detection might be better here in case there are intentional fall-through cases where more than 1 condition might result in the same case. Just a suggestion.

elavoie:
I had not seen this pull-request yesterday, mine (https://github.com/feross/simple-peer/pull/125) might become redundant.

index.js
feross: There aren't any of those cases though. Thanks for the suggestion

PR #125, opened 2017-01-17
Fixed incorrect getStats usage with wrtc

elavoie:
The wrtc getStats signature does not conform to the expected fallback. See https://github.com/js-platform/node-webrtc/blob/develop/lib/peerconnection.js.

Added another option to find whether we are using wrtc and use the wrtc conventions in that case.


feross:
Thanks, but like you mentioned, this is redundant with my PR here: https://github.com/feross/simple-peer/pull/123 Thanks for your effort!

PR #127, opened 2017-01-18
remove debugging console.log leftovers

diasdavid:
@feross, tried the latest version and works like a charm, thank you :) 

I found that there were 3 console.logs left, probably leftovers from testing


feross:
6.2.1

Issue #145, opened 2017-02-20
Browser differences in constraint objects


RationalCoding:
There are currently two ways to specify answer/offer constraints:

The new format:
```javascript
{
  offerToReceiveVideo: true,
  offerToReceiveAudio: true
}
```

The old format:
```javascript
{
  mandatory: {
    OfferToReceiveVideo: true,
    OfferToReceiveAudio: true
  }
}
```

Chrome doesn't support the new spec as of v56, throwing a "malformed constraints object error".

Firefox supposedly supports both formats, but there are a number of issues with using the old format (like onaddstream not firing in certain situations).

It would be great if simple-peer could abstract this away.

feross:
Agreed!

PR #146, opened 2017-02-20
Transform constraints objects

RationalCoding:
Implements #145 

Added a function that transforms constraint objects into either the old or new formats.

Unfortunately browser detection is the only way I can think of to do this... currently detects Chromium (the only browser that needs the old format) via `window.chrome`. Any objections?

### DIFF
--- a/index.js
+++ b/index.js
@@ -35,9 +35,9 @@ function Peer (opts) {
   self.initiator = opts.initiator || false
   self.channelConfig = opts.channelConfig || Peer.channelConfig
   self.config = opts.config || Peer.config
-  self.constraints = transformConstraints(opts.constraints || Peer.constraints)
-  self.offerConstraints = transformConstraints(opts.offerConstraints || {})
-  self.answerConstraints = transformConstraints(opts.answerConstraints || {})
+  self.constraints = self._transformConstraints(opts.constraints || Peer.constraints)
+  self.offerConstraints = self._transformConstraints(opts.offerConstraints || {})
+  self.answerConstraints = self._transformConstraints(opts.answerConstraints || {})
   self.reconnectTimer = opts.reconnectTimer || false
   self.sdpTransform = opts.sdpTransform || function (sdp) { return sdp }
   self.stream = opts.stream || false
@@ -81,6 +81,7 @@ function Peer (opts) {
   // possible for certain implementations.
   self._isWrtc = Array.isArray(self._pc.RTCIceConnectionStates)
   self._isReactNativeWebrtc = typeof self._pc._peerConnectionId === 'number'
+  self._isChromium // _isChromium is needed eariler, see _transformConstraints
 
   self._pc.oniceconnectionstatechange = function () {
     self._onIceConnectionStateChange()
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -683,4 +683,64 @@ Peer.prototype._debug = function () {
   debug.apply(null, args)
 }
 
+// Transform constraints objects into the new format (unless Chromium)
+// This can be removed when Chromium supports the new format
+function transformConstraints (constraints) {
+  if (typeof (window) === 'undefined' || Object.keys(constraints).length === 0) {
```
feross: Can you remove the parens around `window`? They're not needed.

index.js
```diff
@@ -683,4 +683,64 @@ Peer.prototype._debug = function () {
...
+    return constraints
+  }
+  var isChromium = window.chrome
```
feross: Can you move this to line 83, near the other user agent sniffing code? https://github.com/feross/simple-peer/blob/master/index.js#L83 It's better for all the user agent sniffing hacks to be in one place, IMO.

index.js
```diff
@@ -683,4 +683,64 @@ Peer.prototype._debug = function () {
...
+      for (key in constraints.optional) {
+        newConstraints[key] = constraints.optional[key]
+      }
+    }
+
+    if (constraints.mandatory) {
+      for (key in constraints.mandatory) {
+        newConstraints[key] = constraints.mandatory[key]
+      }
+    }
```
feross: I think this code can be simplified by using `Object.assign()`:

```js
var newConstraints = Object.assign({}, constraints.optional, constraints.mandatory)
```

index.js
```diff
@@ -683,4 +683,64 @@ Peer.prototype._debug = function () {
   debug.apply(null, args)
 }
 
+// Transform constraints objects into the new format (unless Chromium)
+// This can be removed when Chromium supports the new format
```
feross: Can you prefix this comment with "TODO" so that it's searchable? I periodically check for "TODO" before publishing new versions and see what can be improved because of new browser releases :)

index.js
```diff
@@ -683,4 +683,64 @@ Peer.prototype._debug = function () {
...
+      delete newConstraints['OfferToReceiveVideo']
+    }
+
+    if (newConstraints.OfferToReceiveAudio !== undefined) {
+      newConstraints.offerToReceiveAudio = newConstraints.OfferToReceiveAudio
+      delete newConstraints['OfferToReceiveAudio']
+    }
+
+    return newConstraints
+  } else if (!(constraints.mandatory || constraints.optional) && isChromium) {
```
feross: I think this would be clearer if written as:

```js
!constraints.mandatory && !constraints.optional && isChromium
```

index.js
RationalCoding: isChromium is needed before the constraints are transformed, but the other user agent sniffing is done after the pc is made. I can't them in the same block, but I'll move it up into the constructor.

RationalCoding:
Great! The differences were causing me some difficulty with that other issue, so glad to have it fixed.

Also, sure. I use simple-peer quite a bit now, so I'm sure I'll be stopping in again.

feross:
Published as 6.4.0, and added you as a collaborator.

feross:
Actually, 6.4.0 isn't published yet, as the tests are stalling. I think we forgot some `peer.destroy()` calls.

feross:
Also now getting this error:

screen shot 2017-02-20 at 4 11 54 am

Sorry, should have looked at the CI test results before merging.

RationalCoding:
I messed up the conditional while refactoring. My mistake, fixing it now.

feross:
It also looks like constraints are no longer passed into the `RTCPeerConnection` object, only to `createOffer` and `createAnswer`. Spec link here: https://w3c.github.io/webrtc-pc/#rtcpeerconnection-interface

I think we should stop passing the second argument into the `RTCPeerConnection` constructor and instead treat `constraints` as a default that is used when the more specific `offerConstraints` or `answerConstraints` are not specified. What do you think?

feross:
Re: my previous comment, I sent a PR here: https://github.com/feross/simple-peer/pull/150

Issue #143, opened 2017-02-19
Event listener of a created instance with stream doesn't work on firefox and edge browser


infocentre:
navigator.mediaDevices.getUserMedia({ audio: false, video: true }).then(function(stream) {
  var peer = new SimplePeer({ initiator: true, stream: stream })
  console.log(peer)
  peer.on('signal', function(data){
    console.log('printing.....')
  });
})

it is working well without stream:stream, and chrome. but on firefox it doesnt recognize event listener .on
so console.log doesn't  say 'printing....'  only not working on firefox, edge. on chrome it works well

RationalCoding:
It looks like `self._pc.onnegotiationneeded` isn't being fired on Firefox when `initator: true`. I think this a Firefox regression although I can't find anything on it.

The hack used for wrtc works for this, so it could be applied to Firefox too? It doesn't seem to have any ill effects on Chrome either.
```javascript
    if (self._isWrtc) {
      self._pc.onnegotiationneeded()
    }
```

@infocentre Until there's a fix, you can manually fire that event like this:
```javascript
navigator.mediaDevices.getUserMedia({ audio: false, video: true }).then(function(stream) {
  var peer = new SimplePeer({ initiator: true, stream: stream })
  console.log(peer)
  peer.on('signal', function(data){
    console.log('printing.....')
  });
  peer._pc.onnegotiationneeded()
})
```

feross:
We used to manually call `self._pc.onnegotiationneeded()` for Firefox, but I discovered it wasn't needed anymore since Firefox has finally implemented the event. Our mistake is that we're calling `addStream()` before we attach the 'negotiationneeded' handler, so it's firing before we're listening for it.

feross:
Fixed in 6.4.1

PR #147, opened 2017-02-20
Listen to 'negotiationneeded' before calling pc.addStream()

feross:
Fixes https://github.com/feross/simple-peer/issues/143

We used to manually `call self._pc.onnegotiationneeded()` for Firefox, but I discovered it wasn't needed anymore since Firefox has finally implemented the event. So I removed that call for Firefox, leaving it only for `wrtc`. The mistake I made is that we're calling `addStream()` before we attach the 'negotiationneeded' handler, so it's firing before we're listening for it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -92,6 +92,22 @@ function Peer (opts) {
     self._onIceCandidate(event)
   }
 
+  if (self.initiator) {
+    var createdOffer = false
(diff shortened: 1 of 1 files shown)


feross:
Thanks for the review, @RationalCoding. I appreciate it.

PR #149, opened 2017-02-20
Use addTrack() instead of addStream() in supported browsers

feross:
`addStream()` is deprecated, so let's use `addTrack()` if the browser
supports it.

Docs that show the deprecation: https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addStream

### DIFF
--- a/index.js
+++ b/index.js
@@ -111,15 +111,19 @@ function Peer (opts) {
     }
   }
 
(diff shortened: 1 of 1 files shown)


feross:
Thanks for all the reviews, @RationalCoding!

PR #150, opened 2017-02-20
Remove second argument to RTCPeerConnection

feross:
It looks like constraints are no longer passed into the
`RTCPeerConnection` object, only to `createOffer` and `createAnswer`.
Spec link here:
https://w3c.github.io/webrtc-pc/#rtcpeerconnection-interface

I think we should stop passing the second argument into the
`RTCPeerConnection` constructor and instead treat `constraints` as a
default that is used when the more specific `offerConstraints` or
`answerConstraints` are not specified.

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,15 +32,10 @@ function Peer (opts) {
     ? opts.channelName || randombytes(20).toString('hex')
     : null
 
-  // Needed by _transformConstraints, so set this early
-  self._isChromium = typeof window !== 'undefined' && window.chrome
 
   self.initiator = opts.initiator || false
   self.channelConfig = opts.channelConfig || Peer.channelConfig
   self.config = opts.config || Peer.config
-  self.constraints = self._transformConstraints(opts.constraints || Peer.constraints)
-  self.offerConstraints = self._transformConstraints(opts.offerConstraints || {})
-  self.answerConstraints = self._transformConstraints(opts.answerConstraints || {})
   self.reconnectTimer = opts.reconnectTimer || false
   self.sdpTransform = opts.sdpTransform || function (sdp) { return sdp }
   self.stream = opts.stream || false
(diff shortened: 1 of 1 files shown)


RationalCoding:
The behaviour is the same either way and this cleans the code up a bit, so I agree.

feross:
@RationalCoding Can you approve the PR with the GitHub review tool, so I can get a green merge button? 😉

RationalCoding:
Any idea why the Travis build failed on my test? Runs fine locally.

feross:
@RationalCoding Looks like the tests that use the `wrtc` implementation of WebRTC are failing:

```
# old constraint formats are used
events.js:166
      throw err;
      ^
Error: Uncaught, unspecified "error" event. ([object Object])
    at Peer.emit (events.js:164:17)
    at Peer._destroy (/home/travis/build/feross/simple-peer/index.js:293:17)
    at Peer._onError (/home/travis/build/feross/simple-peer/index.js:681:8)
    at /home/travis/build/feross/simple-peer/index.js:408:28
    at Daemon.daemon.once (/home/travis/build/feross/simple-peer/node_modules/electron-webrtc/src/RTCPeerConnection.js:274:11)
    at Object.onceWrapper (events.js:291:19)
    at emitOne (events.js:96:13)
    at Daemon.emit (events.js:189:7)
    at ChildProcess.<anonymous> (/home/travis/build/feross/simple-peer/node_modules/electron-eval/lib/index.js:195:27)
    at emitOne (events.js:96:13)
```

You can run them locally with: `WRTC=electron-webrtc tape test/*.js`

RationalCoding:
Yeah they're failing locally as well, but only with these changes.

RationalCoding:
It looks like `wrtc` throws an error when `answerConstraints` are set.

RationalCoding:
Since all implementations still support the second argument in `RTCPeerConnection`, we might be better off leaving this as is instead of changing behaviour for `wrtc`.

feross:
Sounds good.

PR #152, opened 2017-02-23
Only fire 'stream' once per pair of tracks

RationalCoding:
Since 'onTrack' fires the 'stream' event, MediaStreams with both audio and video will fire it twice. (See this [CodePen](https://codepen.io/RationalCoding/pen/JWPeXZ?editors=1111))

This PR stores the IDs so that 'stream' will only be fired on the first track.


index.js
```diff
@@ -111,6 +111,7 @@ function Peer (opts) {
     }
   }
 
+  self._previousStreams = []
```
feross: This should be higher up, next to where the other private variables are declared, IMO.

index.js
```diff
@@ -676,6 +677,9 @@ Peer.prototype._onTrack = function (event) {
   var self = this
   if (self.destroyed) return
   self._debug('on track')
+  var id = event.streams[0].id
+  if (self._previousStreams.indexOf(id) !== -1) return // Only fire one 'stream' event per track
+  self._previousStreams.push(id)
```
feross: We should set `self._previousStreams` to `null` in `destroy()`.

index.js
```diff
@@ -676,6 +677,9 @@ Peer.prototype._onTrack = function (event) {
   var self = this
   if (self.destroyed) return
   self._debug('on track')
+  var id = event.streams[0].id
+  if (self._previousStreams.indexOf(id) !== -1) return // Only fire one 'stream' event per track
```
feross: I think this comment could be clearer. How about "Only fire one 'stream' event, even though there may be multiple tracks per stream"

RationalCoding:
Added the fix for #153

feross:
Released as 6.4.4

PR #172, opened 2017-04-13
Fix channelName assign bug

maxbaluev:
An incorrect name was assigned if the instance was not an initiator.

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,7 +28,7 @@ function Peer (opts) {
 
   stream.Duplex.call(self, opts)
 
-  opts.channelName ? self.channelName = opts.channelName : randombytes(20).toString('hex')
+  opts.channelName ? self.channelName = opts.channelName : self.channelName = randombytes(20).toString('hex')
 
   // Needed by _transformConstraints, so set this early
   self._isChromium = typeof window !== 'undefined' && !!window.webkitRTCPeerConnection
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -28,7 +28,7 @@ function Peer (opts) {
 
   stream.Duplex.call(self, opts)
 
-  opts.channelName ? self.channelName = opts.channelName : randombytes(20).toString('hex')
+  opts.channelName ? self.channelName = opts.channelName : self.channelName = randombytes(20).toString('hex')
```
RationalCoding: This is better written as `self.channelName = opts.channelName || randombytes(20).toString('hex')`

RationalCoding:
@feross Is there a reason the channelName is set to null when `initiator=false`?

maxbaluev:
I have error in tests: `Error:
Zuul tried to run tests in saucelabs, however no saucelabs credentials were provided.
See the zuul wiki (https://github.com/defunctzombie/zuul/wiki/Cloud-testing) for info on how to setup cloud testing.`

feross:
@RationalCoding Are you saying that this line does nothing?

https://github.com/feross/simple-peer/blob/00d6b40f3758ffe26551bfae6fa18a9b17edf01f/index.js#L330

The reason it's set to `null` is that we want the channel name on the non-initiator to match the name that the initiator set, so it must be `null` until we get a 'datachannel' event and can read the label off the datachannel object.

It's better to set it to `null` until we know the true value, since it's explicit: we don't know the value yet.

feross:
@maxbaluev Thanks for the PR, but I'm going to close this for the reasons I mentioned above.

RationalCoding:
@feross I missed that line, but yes - that's what I would expect to happen.

PR #173, opened 2017-04-13
Fix signal value in createOffer/Answer

MatthieuLemoine:
Hi!

In ``_createOffer`` and ``_createAnswer``, ``signal`` should be assign the value of ``offer`` instead of ``pc.localDescription || offer`` because ``setLocalDescription`` is async.

It works the first time as ``pc.localDescription`` is ``undefined``. But if a reoffer happens, the previous value of ``localDescription`` will be used and sent to the remote peer.

### DIFF
--- a/index.js
+++ b/index.js
@@ -401,7 +401,7 @@ Peer.prototype._createOffer = function () {
     offer.sdp = self.sdpTransform(offer.sdp)
     self._pc.setLocalDescription(offer, noop, function (err) { self._destroy(err) })
     var sendOffer = function () {
-      var signal = self._pc.localDescription || offer
+      var signal = offer
       self._debug('signal')
       self.emit('signal', {
         type: signal.type,
@@ -425,7 +425,7 @@ Peer.prototype._createAnswer = function () {
     else self.once('_iceComplete', sendAnswer)
 
     function sendAnswer () {
-      var signal = self._pc.localDescription || answer
+      var signal = answer
       self._debug('signal')
       self.emit('signal', {
         type: signal.type,


feross:
I see your point, but this change would prevent the user from changing the local description. It's decently common for users to create an offer, then "munge" the SDP with a regex before calling `setLocalDescription`. So we need to continue supporting that.

feross:
Actually, sorry that's not correct. We have the `sdpTransform` option which allows the user to munge the SDP before it's set as the local description.

The true reason for the `self._pc.localDescription || offer` line is explained by this commit: https://github.com/feross/simple-peer/commit/3618ba416d8546343eccf1ddeb11946451f72098

This line was added to support the `trickle: false` option, which disables "trickle ICE" (for details on what that is, see: https://webrtchacks.com/trickle-ice/).

In the normal case (`trickle: true`), a 'signal' event is emitted when we have an offer, and again for each ICE candidate that is discovered.

In this other case (`trickle: false`), a single 'signal' event is emitted only after all ICE candidates are gathered. This single object therefore must contain the offer and all the ICE candidates. So, for that reason, we can't just use the `offer` from `createOffer`. We need to get the full description from `self._pc.localDescription` which contains both the offer and the list of ICE candidates.

feross:
Just checked the Travis test output and it stalls on the "disable trickle" tests, so this confirms my analysis.

feross:
I think the right solution here is actually pretty straightforward. Let's just wait until the async `setLocalDescription` function returns success before possibly using the value of `self._pc.localDescription`.

PR here: https://github.com/feross/simple-peer/pull/174

feross:
Thanks for the PR, @MatthieuLemoine. I'll close this PR now. Cheers!

RationalCoding:
This was blocking "trick:false" during renegotiation too. Great!

feross:
Released as 8.1.0.

Issue #178, opened 2017-04-28
Weird behavior in Adjacent Chrome Tabs


carloslfu:
Hi, I have written [an example](https://carloslfu.github.io/simple-peer-bug/app.html) and the [source code is here](https://github.com/carloslfu/simple-peer-bug), these are the steps for reproduce:

- Open [example](https://carloslfu.github.io/simple-peer-bug/app.html) in two adjacent Chrome Tabs (I am using Linux Mint 18) and Open the JS console in both. Adjacent Tabs means both are in the same window
- Copy `Your Offer` from first Tab to `Remote Offer` in second Tab and press Save
- Copy the answer from second Tab to first Tab `Remote Answer` input and press Save button

Expected:

- JS console should show `CONNECTED` in both Tabs

Actual:

- Only the offerer Tab show `CONNECTED` message

Note that if you probe it with Tabs in different windows this works as expected.

matthewtoast:
My work around, FWIW: Once the offerer (initiator) hears the `connected` event, have it begin sending a regular 'heartbeat' message. The non-offerer (guest) will begin receiving this data despite never having fired its own `connected` event (it is connected despite not saying so!). On the non-offerer (guest) side, treat the first 'heartbeat' you receive as a substitute for the `connected` event.

carloslfu:
Thanks @matthewtoast

SupremeTechnopriest:
I'm also experiencing the same issue. However, I'm running two entirely separate chrome instances:

`/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --user-data-dir=/tmp/XXX`

...which leads me to believe this problem goes further than just neighboring tabs.  I know chrome 58 was just released and I see these todos ([1](https://github.com/feross/simple-peer/blob/master/index.js#L599), [2](https://github.com/feross/simple-peer/blob/master/index.js#L616)) in the source code, but I don't suspect it's related.

RationalCoding:
Seems `self._channel.onopen` is not being fired when connecting two tabs in the same window.

What is strange is that `self._channel.onmessage` is still being fired, hence why @matthewtoast 's solution works.

Created a PR to always fire "connect" before emitting data. This should allow you to use normal data to determine connection, instead of sending many heartbeats (assuming you send data on connect on the other peer).

This is definitely a Chrome bug, and we won't have a real solution until they fix it.

SupremeTechnopriest:
[...]

PR #180, opened 2017-06-01
Workaround for #178

RationalCoding:
Forces the 'connect' event to always fire before any 'data' events (or similar).

The 'connect' event still won't fire if no data is sent, but removes the need to constantly send "heartbeats".


RationalCoding:
@feross Tests are only failing due to saucelabs.

feross:
Released as 8.1.1

PR #185, opened 2017-06-16
Error processing ICE Candidate

cdvv7788:
I was getting an error with processing ICE Candidate.
I fixed it following https://stackoverflow.com/questions/38198751/domexception-error-processing-ice-candidate
Working as expected for me now.

I suspect that it was failing when one of the peers was trying to renegotiate and the candidates for the previous connection were still being sent.
Anyway, this is missing a detail according to the solution provided on SO, which is to push the candidates to a queue. However it works and i am not getting any issue on my connections after this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -216,14 +216,17 @@ Peer.prototype.signal = function (data) {
 
 Peer.prototype._addIceCandidate = function (candidate) {
   var self = this
-  try {
-    self._pc.addIceCandidate(
-      new self._wrtc.RTCIceCandidate(candidate),
-      noop,
-      function (err) { location.reload(); self._destroy(err) }
-    )
-  } catch (err) {
-    self._destroy(new Error('error adding candidate: ' + err.message))
+  if(!self._pc || !self._pc.remoteDescription.type){}
+  else{
+    try {
+      self._pc.addIceCandidate(
+        new self._wrtc.RTCIceCandidate(candidate),
+        noop,
+        function (err) { self._destroy(err) }
+      )
+    } catch (err) {
+      self._destroy(new Error('error adding candidate: ' + err.message))
+    }
   }
 }
 
--- a/index.js
+++ b/index.js
@@ -216,8 +216,10 @@ Peer.prototype.signal = function (data) {
 
 Peer.prototype._addIceCandidate = function (candidate) {
   var self = this
-  if(!self._pc || !self._pc.remoteDescription.type){}
-  else{
+  if (!self._pc || !self._pc.remoteDescription.type) {
+    // Add candidate to queue
+  }
+  else {
     try {
       self._pc.addIceCandidate(
         new self._wrtc.RTCIceCandidate(candidate),
--- a/index.js
+++ b/index.js
@@ -218,8 +218,7 @@ Peer.prototype._addIceCandidate = function (candidate) {
   var self = this
   if (!self._pc || !self._pc.remoteDescription.type) {
     // Add candidate to queue
-  }
-  else {
+  } else {
     try {
       self._pc.addIceCandidate(
         new self._wrtc.RTCIceCandidate(candidate),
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -216,14 +216,17 @@ Peer.prototype.signal = function (data) {
...
   var self = this
-  try {
-    self._pc.addIceCandidate(
-      new self._wrtc.RTCIceCandidate(candidate),
-      noop,
-      function (err) { self._destroy(err) }
-    )
-  } catch (err) {
-    self._destroy(new Error('error adding candidate: ' + err.message))
+  if(!self._pc || !self._pc.remoteDescription.type){}
```
RationalCoding: The candidate should be pushed to a queue here, if your solution is correct.

index.js
```diff
@@ -220,7 +220,7 @@ Peer.prototype._addIceCandidate = function (candidate) {
     self._pc.addIceCandidate(
       new self._wrtc.RTCIceCandidate(candidate),
       noop,
-      function (err) { self._destroy(err) }
+      function (err) { location.reload(); self._destroy(err) }
```
RationalCoding: simple-peer also runs in Node, so location isn't appropriate here. Reload should be handled externally.

cdvv7788: It was a temporary workaround. I removed it in the next commit.

index.js
cdvv7788: Yeah, i agree. However I am pretty new to WebRTC and this package and I don't want to mess the application flow.
Can you give me some hints or help me with that part?

cdvv7788:
@RationalCoding  The travis check is raising an error: Zuul tried to run tests in saucelabs, however no saucelabs credentials were provided.
Something I am missing?

RationalCoding:
It's not your fault. You can run the browser tests locally with `npm run test-browser-local`. They seem to be passing on Firefox and Chrome, so all's good.

RationalCoding:
@cdvv7788 It seems the ice candidates are already pushed to a queue if the remote description isn't set: see line 196-199.

```javascript
  if (data.candidate) {
    if (self._pc.remoteDescription) self._addIceCandidate(data.candidate)
    else self._pendingCandidates.push(data.candidate)
  }
```

ICE failures are intermittent, and the apparent fix might have been a coincidence... I don't see any reason to add another queue.

cdvv7788:
@RationalCoding yet it fixed the issue for me. I went from having issues connecting 2 peers to reliably being able to connect 4 (only adding the condition, not the queue).

RationalCoding:
So you can reproduce the ice candidate failure with and without the fix? Ok, maybe we have something wrong with the existing condition then.

RationalCoding:
Could you try `self._pc.remoteDescription` instead of `!self._pc || !self._pc.remoteDescription.type` and see if that still works?

cdvv7788:
I can reliable reproduce:
Without fix i get, intermittently: DOMException: Error processing ICE candidate
Takes around 10 refreshes to reproduce.

With the fix i cannot reproduce that issue. Tried 100 times without success.

index.js
```diff
@@ -216,14 +216,18 @@ Peer.prototype.signal = function (data) {
...
-  try {
-    self._pc.addIceCandidate(
-      new self._wrtc.RTCIceCandidate(candidate),
-      noop,
-      function (err) { self._destroy(err) }
-    )
-  } catch (err) {
-    self._destroy(new Error('error adding candidate: ' + err.message))
+  if (!self._pc || !self._pc.remoteDescription.type) {
+    // Add candidate to queue
```
RationalCoding: Can you place a breakpoint here to see if this is ever reached? That would suggest the conditions are not identical. I would, but I can't seem to reproduce ICE failures at the moment.

cdvv7788: I can. It is being entered. Do you want a dump of the peer or something?

RationalCoding:
It was indeed the condition (but only with certain STUN/TURN servers, which explains why I couldn't reproduce. Are you using coturn?)

I've refactored your PR as #210 so that we can get this merged quickly. Thanks again.

cdvv7788:
Yeah, I am using coturn.

PR #210, opened 2017-11-03
Fix ice candidates being added before valid remoteDescription is set

RationalCoding:
Refactor of #185.

When the remoteDescription is set, it can sometimes still have an empty type field, making the check a little more complex. My guess is incorrect SDP information, but I think this is preferable to munging the SDP lines.

Incorrect Ice Candidate errors (which I've only noticed with coturn servers) vanish when this fix is applied.

Thanks to @cdvv7788 for the initial PR.

### DIFF
--- a/index.js
+++ b/index.js
@@ -195,7 +195,7 @@ Peer.prototype.signal = function (data) {
   self._debug('signal()')
 
   if (data.candidate) {
-    if (self._pc.remoteDescription) self._addIceCandidate(data.candidate)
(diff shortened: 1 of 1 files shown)


RationalCoding:
Are there any active contributors that are able to review this and the other PRs? I'm not able to merge my own PRs (and would like a second opinion regardless).

AlbertoElias:
Would be great to get this merged!

RationalCoding:
@feross While you're around, do you mind reviewing this fix? CoTurn servers have a lot of problems without it.

PR #215, opened 2017-11-28
Fix partial incompatibility with wrtc 0.0.63

nazar-pc:
Fix partial incompatibility with wrtc 0.0.63 that doesn't guarantee to have candidate pair with active connection when RTCDataChannel is opened (see https://github.com/js-platform/node-webrtc/issues/339 for details)

Test is added in order to reproduce the issue and confirm the fix. Easier to review with `?w=1` is URL in order to hide code indentation changes.

cc @markandrus


markandrus:
@nazar-pc this looks like it would work to me; however, I guess you had to introduce the `withinIntervalCallback` because there is no guarantee that a `getStats` request will resolve before the next interval. Is that right? If so, would it be simpler to eliminate `withinIntervalCallback` and to use `setTimeout` instead of `setInterval` to re-queue a `getStats` request if we haven't found the candidate pair yet?

nazar-pc:
Great suggestion, I've made it even simpler now

markandrus:
Nice! 👍 LGTM

.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
RationalCoding: Please don't commit this file (git exclude).

index.js
```diff
@@ -547,120 +550,132 @@ Peer.prototype._maybeReady = function () {
...
+        }
+      })
 
-    items.forEach(function (item) {
-      // TODO: Once all browsers support the hyphenated stats report types, remove
-      // the non-hypenated ones
-      if (item.type === 'remotecandidate' || item.type === 'remote-candidate') {
-        remoteCandidates[item.id] = item
+      function setSelectedCandidatePair (selectedCandidatePair) {
+        self._connecting = false
```
RationalCoding: The use of `self._connecting` and `self.connected` like this means that `self.connected` will be false on a getStats error. These should be left near where the events are emitted, and a different variable can be used.

index.js
```diff
@@ -547,120 +550,132 @@ Peer.prototype._maybeReady = function () {
...
-        // Firefox
-        self.remoteAddress = remote.ipAddress
-        self.remotePort = Number(remote.portNumber)
-      } else if (typeof selectedCandidatePair.googRemoteAddress === 'string') {
-        // TODO: remove this once Chrome 58 is released
-        remote = selectedCandidatePair.googRemoteAddress.split(':')
-        self.remoteAddress = remote[0]
-        self.remotePort = Number(remote[1])
+      self._debug('connect')
+      self.emit('connect')
```
RationalCoding: This event can fire without `self.connected` being set to true if items is empty.

.gitignore
nazar-pc: Removed, but why? It will be present in literally any clone and `node_modules` definitely shouldn't appear in repository.

nazar-pc:
Applied some tweaks, does it look better now?

RationalCoding:
Seems okay to me. It's unfortunate wrtc requires this hack, but I don't see it getting fixed any time soon.

nazar-pc:
I don't think it will ever be fixed at all. As mentioned in linked issue latest Chrome also seems to have the same behavior and spec doesn't guarantee anything in this regard anyway.

nazar-pc:
When to expect new release containing this fix?

feross:
8.2.0

PR #222, opened 2018-01-02
wrtc@0.0.65 supports sending Buffers

### LINKED ISSUE #60: Sending buffers fails over wrtc sometimes
When I run the unit tests for node.js, they fail:

```
# data send/receive Buffer
isbuffer true true
ok 24 data is Buffer
not ok 25 got correct message
  ---
    operator: deepEqual
    expected: |-
      <Buffer >
    actual: |-
      <Buffer 76 61 72 20 63 6f 6d 6d 6f 6e 20 3d 20 72 65 71 75 69 72 65 28 27 2e 2f 63 6f 6d 6d 6f 6e 27 29 0a 76 61 72 20 50 65 65 72 20 3d 20 72 65 71 75 69 72 ... >
    at: Peer.<anonymous> (/Users/paul/repos/simple-peer/test/binary.js:71:9)
  ...
isbuffer true true
FATAL ERROR: v8::ArrayBuffer::Externalize ArrayBuffer already externalized
^C
```

Bizarrely, that Buffer translates to:

```
var common = require('./common')
var Peer = requir...
```

I think the reason it's failing is because simple-peer isn't converting Buffers to UInt8Arrays, because isTypedArray returns true for... most Buffers? All Buffers? I'll try to make a pull request that fixes this.

markandrus:
Fixes #60


RationalCoding:
One less hack. Thanks! :)

feross:
Nice!

feross:
Released as 8.4.0

PR #228, opened 2018-02-01
Fix for Safari 11

nazar-pc:
Safari 11 implementation looks spec-compliant, but it currently lacks `transport` items as well as `local-candidate` and `remote-candidate`, let connection finish even if local/remote candidates are not available


RationalCoding:
[This test fails](https://github.com/feross/simple-peer/blob/1650fefbfdf700f62be717463acc8f958a0d0f6b/test/basic.js#L179).

Does Safari not support remoteAddress and remotePort?

nazar-pc:
Safari doesn't provide this information yet unfortunately, so test is expected to fail in Safari.

RationalCoding:
Looks good. Thanks for debugging this.

feross:
Released as 8.4.0

PR #232, opened 2018-02-17
make constructor in examples consistent

feross:
SimplePeer -> Peer


RationalCoding:
Nice, this was a bit confusing to people who aren't used to Browserify.

PR #236, opened 2018-02-17
Add Sauce Labs tests for Desktop Safari and iOS 

feross:
Fixes: #227

I had to skip some test suites on Safari + iOS to get it to pass, specifically these ones:

- disable trickle
- disable trickle (only initiator)
- disable trickle (only receiver)
- ensure remote address and port are available right after connection

Would be nice to fix these up eventually and remove the code to skip.

### DIFF
--- a/.zuul.yml
+++ b/.zuul.yml
@@ -4,3 +4,7 @@ browsers:
     version: latest
   - name: chrome
     version: latest
+  - name: safari
+    version: latest
+  - name: iphone
+    version: latest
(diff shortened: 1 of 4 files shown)


feross:
It looks like there's issues with all the "disable trickle" variant tests. I'm going to disable them all in Safari + iOS and merge this. I want to test what is currently working in Safari and iOS so that we don't regress.

We can fix the skipped tests in Safari + iOS in a future PR

RationalCoding:
The "ensure remote address and port" simply isn't supported on Safari. I'll look into what's happening with disabling trickle.

Issue #187, opened 2017-06-29
Node v8 stream destroy and _destroy methods


mwiktorczyk:
Node v8 introduced some changes to stream API. One of them is `destroy` method:

- https://nodejs.org/api/stream.html#stream_readable_destroy_error
- https://nodejs.org/api/stream.html#stream_writable_destroy_error

According to this specification, 
> implementors should not override this method, but instead implement readable._destroy

Is there a chance to adapt your library to node v8 and later on fix the TypeScript definitions?

RationalCoding:
`.destroy` simply calls `._destroy`, so is this as simple as removing `Peer.prototype.destroy`?

RationalCoding:
`readable-stream` doesn't seem to implement the same behaviour. (calling `.destroy` doesn't call `._destroy`)

It seems it would be as simple as removing `._destroy`.

One thing to consider is that the Node V8 `.destroy` method isn't documented as having a callback parameter, but takes an error.

feross:
I think we can switch to the new interface and just shim the missing `.destroy()` method until this issue is fixed: https://github.com/nodejs/readable-stream/issues/283

This is a breaking change since it changes the API, but it makes the API match Node.js which is aligns with user expectations and the stated goal of this package (to emulate `net.Socket` where possible).

PR here: https://github.com/feross/simple-peer/pull/237

PR #237, opened 2018-02-17
 BREAKING: Change .destroy(onclose) to .destroy(err)

feross:
Node.js 8 standardized the destroy() stream interface, so let's use that interface.

To update your code, change this:

```js
peer.destroy(callback)
```

To this instead:

```js
peer.on('close', callback)
peer.destroy()
```

Fixes: #187


feross:
Released as 9.0.0.

PR #241, opened 2018-02-20
Replace zuul with airtap

feross:
`airtap` is actually maintained and much simpler (just meant for `tape`) ✨

### DIFF
--- a/.airtap.yml
+++ b/.airtap.yml
@@ -1,4 +1,5 @@
-ui: tape
+sauce_connect: true
+loopback: airtap.local
 browsers:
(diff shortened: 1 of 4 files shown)


feross:
Looks like the `electron-webrtc` tests are failing now, specifically in the 'cleanup' test. https://travis-ci.org/feross/simple-peer/builds/342642745

I confirmed that the tests are failing in the same place on `master` now too, so this wasn't made worse by switching to airtap. Unclear upon quick inspection what the problem is. But it may have to do with a transitive dependency getting updated...

Issue #238, opened 2018-02-18
Meta information about error context passed to _destroy and .emit('error', x)


joshontheweb:
It would be helpful to have some meta information about the source function in which the error happened passed along with error events.  Since the errors all get channeled into the one _destroy function before being emitted, it can be hard to categorize the errors programmatically as to which stage in the signaling process the error originated in.  That is, unless I'm missing something.

For instance, I would like to be able to log errors categorized by the context of the failure like so:
```
peer.on('error', function (err) {
  var errorContext = 'createOffer'
  reportError(peer._pc, errorContext, err)
})
```

As it stands now, it seems difficult to discern the error context for reporting without doing something like triggering every possible error and trying to build a hardcoded index of what could come from what source.  I also imagine that this approach wouldn't be portable between browsers since some of the browser's native error messages are passed directly through.

What I propose would be something like this using createOffer as an example:
```
Peer.prototype._createOffer = function () {
  var self = this
  if (self.destroyed) return

  self._pc.createOffer(function (offer) {
    if (self.destroyed) return
    offer.sdp = self.sdpTransform(offer.sdp)
    self._pc.setLocalDescription(offer, onSuccess, onError)

    function onSuccess () {
      if (self.destroyed) return
      if (self.trickle || self._iceComplete) sendOffer()
      else self.once('_iceComplete', sendOffer) // wait for candidates
    }

    function onError (err) {
      self._destroy(err, null, 'setLocalDescription') // <-- string representing failure context
    }

    function sendOffer () {
      var signal = self._pc.localDescription || offer
      self._debug('signal')
      self.emit('signal', {
        type: signal.type,
        sdp: signal.sdp
      })
    }
  }, function (err) { self._destroy(err, null, 'createOffer') }, self.offerConstraints) // <-- here again
}


Peer.prototype._destroy = function (err, onclose, errContext) { // <-- take errContext as third optional arg
  var self = this
  ...
  if (err) self.emit('error', err, errContext) // <-- pass errContext through to the error event
  self.emit('close')
}
```

By adding the errContext as a third optional argument to _destroy() it should keep things backwards compatible while enabling more descriptive and granular debugging and error tracking.

I'm not married to this particular solution and welcome alternative ideas on this if there is a better way.

Also I'm happy to put in the leg work and make a pull request as long as there is agreement from the maintainers that this is a welcome addition.  Thanks :)

RationalCoding:
[...]

PR #243, opened 2018-02-22
Added error codes to indicate the origin of failures

joshontheweb:
Here my initial stab at implementing the error codes discussed in #238 

I'm not so confident in some of the categorization, specifically the ERR_DATA_CHANNEL feels like a bit of a catch-all and maybe needs to be split up.  I welcome any feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,6 +345,22 @@ Fired when a fatal error occurs. Usually, this means bad signaling data was rece
 
 `err` is an `Error` object.
 
+## error codes
+
+To make debugging and logging errors easier, `Error` objects have an attached error code.  The `err.code` will indicate the origin of the failure and `err.message` will share further details on what went wrong.
+
+### err.code = ERR_CREATE_OFFER
+
+- `ERR_CREATE_OFFER` - The failure occurred in the createOffer function
+- `ERR_CREATE_ANSWER` - The failure occurred in the createAnswer function
+- `ERR_SET_LOCAL_DESCRIPTION` - The failure occurred in the setLocalDescription function
+- `ERR_SET_REMOTE_DESCRIPTION` - The failure occurred in the setRemoteDescription function
+- `ERR_ADD_ICE_CANDIDATE` - The failure occurred in the addIceCandidate function
+- `ERR_ICE_CONNECTION_FAILURE` - There was a failure in the ice connection
+- `ERR_SIGNALING` - There was a signaling failure
+- `ERR_DATA_CHANNEL` - There was a failure establishing or sending on a data channel
+
+
 ## connecting more than 2 peers?
 
 The simplest way to do that is to create a full-mesh topology. That means that every peer
--- a/README.md
+++ b/README.md
@@ -347,17 +347,18 @@ Fired when a fatal error occurs. Usually, this means bad signaling data was rece
 
 ## error codes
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -805,4 +805,21 @@ Peer.prototype._transformConstraints = function (constraints) {
...
+  ERR_CREATE_OFFER: 'ERR_CREATE_OFFER',
+  ERR_CREATE_ANSWER: 'ERR_CREATE_ANSWER',
+  ERR_SET_LOCAL_DESCRIPTION: 'ERR_SET_LOCAL_DESCRIPTION',
+  ERR_SET_REMOTE_DESCRIPTION: 'ERR_SET_REMOTE_DESCRIPTION',
+  ERR_ADD_ICE_CANDIDATE: 'ERR_ADD_ICE_CANDIDATE',
+  ERR_ICE_CONNECTION_FAILURE: 'ERR_ICE_CONNECTION_FAILURE',
+  ERR_SIGNALING: 'ERR_SIGNALING',
+  ERR_DATA_CHANNEL: 'ERR_DATA_CHANNEL'
+}
+
```
nazar-pc: If these are not exported as constants, is this object even needed?
I'd use plain strings instead or export them as `Peer.ERR_*` constants with numeric codes.

RationalCoding: +1 for plain strings. No need for another layer.

joshontheweb: Yeah, I deliberated over that a bit.  My thinking was that I prefer having them all defined somewhere together so there is less chance of 'rogue' error codes popping up without being noticed.  Exporting constants on Peer might be a happy medium.

RationalCoding:
I think the vaugeness of the error codes is fine. They'll be coupled with a message that will distinguish them.

index.js
joshontheweb: Pushed a new commit with string for the error codes.  Is it normal for the build to hang and fail like this?  Tests passed for me locally but they do stop for quite awhile at some points.

nazar-pc: `electron-webrtc` has some issues, this is why it hangs. Not your fault.

README.md
```diff
@@ -345,6 +345,21 @@ Fired when a fatal error occurs. Usually, this means bad signaling data was rece
 
 `err` is an `Error` object.
 
+## error codes
+
+To make debugging and logging errors easier, `Error` objects have an attached error code. The `err.code` property will indicate the origin of the failure and `err.message` will share further details on what went wrong.
+
+- `ERR_WEBRTC_SUPPORT` - The failure occurred in the webRTC support check
```
RationalCoding: The codes are pretty self-explanatory. I don't think this list is necessary.

nazar-pc: It is sometimes useful albeit somewhat tedious to maintain, to have a list of codes in readme rather than searching through source code.

RationalCoding: A simple list of codes might be useful, but these descriptions don't really add any explanation.

joshontheweb: I'm don't have strong feelings on this.  However the readme section was requested by @feross in the feature request thread (https://github.com/feross/simple-peer/issues/238).

Is there a way I can add more info there to make it more useful?

Happy to remove the section if that is the consensus.

RationalCoding: I've committed suggested changes to the README.

RationalCoding:
LGTM if you are happy with the error code docs. @nazar-pc @joshontheweb

joshontheweb:
@RationalCoding All good from my perspective.

feross:
Nice work!

PR #249, opened 2018-02-26
Remove incorrect JSON documentation

RationalCoding:
We don't parse JSON objects automatically from the datachannel.

I think this was written because we DO automatically parse signal data.


feross:
Actually, we used to automatically do this in a very old version :) That was a bad idea and my mistake!

PR #261, opened 2018-03-19
[WIP] Fix electron-webrtc tests

RationalCoding:
The `onclose` event was not being fired, so two of the tests in the stream suite were failing on electron-webrtc. Oddly enough, I could only reproduce this inside Travis.

Closing the channel before closing the pc seems to make this error disappear. I haven't seen in it the 5 runs since, where I was hitting it every time before.

I think we should figure out why electron-webrtc has this behaviour, but I'm at a loss at how to reproduce outside of Travis.

### DIFF
--- a/index.js
+++ b/index.js
@@ -271,6 +271,16 @@ Peer.prototype._destroy = function (err, cb) {
   if (self._onFinishBound) self.removeListener('finish', self._onFinishBound)
   self._onFinishBound = null
 
+  if (self._channel) {
+    try {
+      self._channel.close()
+    } catch (err) {}
+
+    self._channel.onmessage = null
+    self._channel.onopen = null
+    self._channel.onclose = null
+    self._channel.onerror = null
+  }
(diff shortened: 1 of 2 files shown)


nazar-pc:
Hmm... test is still failing on Travis

RationalCoding:
Now it's throwing ECONNRESET. That's different...

RationalCoding:
The ECONNRESET is thrown by electron-eval abruptly closing the TCP connection to the electron daemon. I'm not sure why this isn't caught by that module considering that's an expected error on closing the daemon, but they seem aware of the issue.

I've silenced the error as it has nothing to do with simple-peer or normal usage of electron-webrtc.

RationalCoding:
electron-webrtc tests pass now. Not sure what is going on with the new airtap tests though.

RationalCoding:
Still seeing ECONNRESET... seems we'll have to fix the problem through electron-eval.

PR #266, opened 2018-03-30
Upgrade wrtc to latest release and drop its detection as unnecessary

nazar-pc:
`_isWrtc` was needed because `onnegotiationneeded` wasn't implemented in wrtc.
Now with wrtc 0.1.0 `onnegotiationneeded` is supported and `onnegotiationneeded` is not longer used after #250 was merged. So there is no reason to keep `_isWrtc`.

Also wrtc starts following semver, hence `^0.1.0`.


feross:
Restarted the tests after merging the airtap update and now the sauce labs tests are skipped because credentials are missing on PRs. :)

feross:
`master` is currently failing (see https://github.com/feross/simple-peer/issues/283) but this PR is passing, so it shouldn't hurt to merge it.

PR #269, opened 2018-04-01
Added Stealthy IM (https://stealthy.im) to the list of projects using…

AC-FTW:
… Simple Peer.


AC-FTW:
Thanks nazar-pc, I've updated the link and will look at our site config.

PR #277, opened 2018-04-10
Always return an object in `signal` event

nazar-pc:
Every other instance of `signal` even emits an object, but this single place for some reason emits JSON string. Let's be consistent.

### DIFF
--- a/index.js
+++ b/index.js
@@ -334,9 +334,9 @@ Peer.prototype.negotiate = function () {
     }
   } else {
     self._debug('requesting negotiation from initiator')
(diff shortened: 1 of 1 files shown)


RationalCoding:
I’d be in favour of this. JSON isn’t always appropriate for signal channels that use a different serialization like MessagePack.

nazar-pc:
This is not about what serialization to use. The point here is that no serialization is needed at all. All other `emit('signal')` use plain objects.

RationalCoding:
Ah, I just realized this is specific to the new negotiation signal event. Good catch.

PR #279, opened 2018-04-11
app addition to who is using simple-peer

akshayKMR:
Hey, I've been using simple-peer in my app [oorja.io](https://oorja.io) - [github](https://github.com/akshayKMR/oorja). Simple peer was super easy to use and its interface is excellent. Thank you developers.


RationalCoding:
Looks great!

Issue #280, opened 2018-04-11
Thoughts on reducing browser bundle size


nazar-pc:
In my current app `simple-peer` is about 10% of an application size and about 20% of JS code size.
Specifically, while `index.js` is 29.5KiB and one might expect that minified version would be even smaller, `simplepeer.min.js` is whopping 86.5KiB right now.

Before I create and start maintaining a fork of this project, I'd like to know if if this is something that can be fixed upstream.

2 packages that cause this huge size:
* `readable-stream`
* `buffer`

`readable-stream` is valuable in Node.js, but not so much in browser environment. Is it possible that future versions of `simple-peer` will stop using it entirely and replace with something lightweight instead? I imagine it wouldn't be too hard to implement a wrapper or even offer an optional one with this library.

`buffer` is a simpler one, once `readable-stream` is eliminated we'll just need to switch from `Buffer` to `Uint8Array`, which is quite easy to do.

Current bundle size doesn't seem tolerable to me given that it is just a wrapper around native features, so I hope for your understanding here. If these changes seem reasonable, I'd be happy to implement them myself, otherwise I'll be forced to create something like `simple-peer-light` and just maintain that fork forever.

nazar-pc:
In fact, I went ahead and removed streaming functionality, converted `Buffer` to `Uint8Array` and got 26.7KiB of minified build. Could be even more savings (about 17KiB) when `debug` module is removed, but even without that an impressive result I think.
https://github.com/feross/simple-peer/compare/master...Detox:8bf7e98e5f354d8f1804d07c63b55dcd2dc9fa46?expand=1

RationalCoding:
Do you think it would be possible to retain the stream API by making a smaller readable-stream alternative? simple-peer is pretty entrenched in the "stream module" ecosystem.

RationalCoding:
We could release two separate bundles (simple-peer and simple-peer-lite), the decoupling wouldn't be too difficult.

The lite version could be built without those larger modules and use a non-streaming `send` API. Replacing a module with an empty stub is fairly simple with browserify and can be detected in the constructor to determine which `send` method to expose. That way we have a single codebase and preserve the existing features.

feross:
Problems with removing it:

- breaking change, all webtorrent packages and many dependents would need to get rewritten
- lots of tests are written against the stream api

We could make a minimal shim and include it in this repo so users could `require('simple-peer/lite')` but I don't want to include a whole second set of tests for the lite version.

This seems like a lot of work.

[...]

PR #312, opened 2018-05-23
Buffer usage elimination #1

nazar-pc:
This is the first step in reducing bundle size as discussed in #280

This PR removes explicit usage of Buffer API in simple-peer itself, but Buffer is still used by `debug` and `readable-stream`, so bundle size will not be much smaller just yet, those will be tacked in a separate PR.

One major implication of this PR is that `data` event will dispatch with `Uint8Array` instead of `Buffer`, which is a major breaking change.

I believe this is a reasonable trade-off since It is easy to upgrade existing code base with `Buffer.from(data)`, while browser users that don't need Buffer API will have significantly less code to download.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,12 +3,24 @@ module.exports = Peer
 var debug = require('debug')('simple-peer')
 var getBrowserRTC = require('get-browser-rtc')
 var inherits = require('inherits')
-var randombytes = require('randombytes')
+var randombytes = require('./lib/randombytes')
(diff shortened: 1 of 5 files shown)


mrlika:
I also disappointed that huge Buffer is used everywhere. Bundles for the browser should be small. Changing the code to use compatible between Node.js and browsers ArrayBuffer/Uint8Array is the way to go.

PR #315, opened 2018-05-25
Update docs to reflect actual STUN URL

pastudan:
Just noticed this was out of date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,7 +214,7 @@ If `opts` is specified, then the default options (shown below) will be overridde
   initiator: false,
   channelConfig: {},
   channelName: '<random string>',
-  config: { iceServers: [ { urls: 'stun:stun.l.google.com:19302' } ] },
+  config: { iceServers: [ { urls: 'stun:23.21.150.121' } ] },
   constraints: {},
   offerConstraints: {},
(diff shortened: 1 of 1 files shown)


pastudan:
Also, I'm just curious why you're not using the google one. Against Google ToS? And are you footing the bill for this STUN server?

nazar-pc:
You have probably looked at the wrong place, here are actual STUN servers used by default:
https://github.com/feross/simple-peer/blob/f3cb23e8fc867a17fc45e38505996d601b90a595/index.js#L151-L160

nazar-pc:
Also readme is just an example, you can place any STUN server you want there or omit altogether and use defaults

pastudan:
Oops! Yes I was looking at an [old commit](https://github.com/feross/simple-peer/commit/822f2ae3b1394d226b18523810d2007b0b5c934c). Sorry about that.

pastudan:
As for my rationale for updating... the readme says these are the defaults. Not a big deal, but its probably good to give people a heads up that by using this package their users will be connecting to Google and Twilio servers.

feross:
@pastudan Thanks, updated the readme.

PR #322, opened 2018-06-10
add new user of simple-peer

lovefoodcode:


RationalCoding:
Nice! I was hoping we'd see some more open-source WebRTC CDNs.

Issue #287, opened 2018-04-13
Implement renegotiation of constraints


RationalCoding:
Currently only renegotiation of media lines is supported. Due to an existing bug (see #95), it's impossible to upgrade a datachannel-only connection to a connection with mediastreams on the non-initiator side. (The initiator can upgrade it's side with no problems.)

Renegotiation of media constraints should be added for this reason and it's generally a good feature to have.

feross:
@RationalCoding Did you link to the wrong issue. I don't think #94 is related, right?

RationalCoding:
Yes, #95.

Issue #95, opened 2016-10-14
stream not added if initiator set to false


dukelec:
Change the video/voice example code from:

```
  var peer1 = new SimplePeer({ initiator: true, stream: stream })
  var peer2 = new SimplePeer({ initiator: false })
```

to:

```
  var peer1 = new SimplePeer({ initiator: false, stream: stream })
  var peer2 = new SimplePeer({ initiator: true })
```

the stream callback never called.

techleadzdevteam:
Hi,
Hope that you will be good.Can you please help me regarding the issue #96

I will be very thankfull to you.

Thanks

RationalCoding:
There's an issue in Chrome with `pc.onaddstream` not firing when the the non-initiator is the one providing the only stream.

The workaround is to specify the offer and answer constraints.

The initiator should offer to receive.  
`new SimplePeer({initiator : true, offerConstraints:{ mandatory: { OfferToReceiveAudio: true, OfferToReceiveVideo: true } }})`

The non-initiator should not offer to receive.  
`new SimplePeer({initiator : false, stream:webcamStream, trickle:false, answerConstraints:{ mandatory: { OfferToReceiveAudio: false, OfferToReceiveVideo: false } }});`

You have to use Chrome's constraint format here, the specs one won't work.

RationalCoding:
There's an open bug with `pc.onaddstream` not firing in Chrome.

https://groups.google.com/forum/#!topic/mozilla.dev.media/BqklQQIXx9w

This forum suggests a workaround with `pc.getRemoteStreams()[0]`, but I haven't been able to get it working.

feross:
@RationalCoding Do you have a link to a Chrome bug? Should this workaround only be applied for Chrome? Any interest in sending a PR to add this workaround to `simple-peer`?

RationalCoding:
[Here's the forum discussing the bug.](https://groups.google.com/forum/#!topic/mozilla.dev.media/BqklQQIXx9w) 

I seem to have lost the forum that provided the workaround, but it definitely solved the issue for me.

The workaround needs to be applied for Chrome, but specifying the constraints shouldn't affect the behaviour of other browsers.

I'm not sure if this can be added to `simple-peer`, since the constraints need to be set with knowledge of what streams will be sent *before* connection.

[...]

PR #324, opened 2018-06-12
Add renegotiation of constraints

RationalCoding:
#287 

Adds the following methods to change constraints. There's not much to them and work the same way as setting constraints in the constructor.

```
peer.setOfferConstraints()
peer.setAnswerConstraints()
peer.setConstraints()
```

Adds tests and documentation for these methods as well. I don't see many people using this except to work around #95 while renegotiating.

I've also fixed an unrelated issue where the non-initiator would send a useless "renegotiate signal" on startup.

Some limitations:

Not all constraints can be arbitrarily changed, as per spec. For example, trying to set `offerToReceiveAudio:false` after you've already accepted an audio track will fail.

Updating constraints on subsequent offers does not work on Safari: https://bugs.webkit.org/show_bug.cgi?id=186576


nazar-pc:
But I have a generic concern about constraints separation. Second argument `constraints` in `RTCPeerConnection` was removed from the spec a long time ago. At the same time answer/offer constraints are obvious from the context.

I'd like to see a single `peer.setConstraints()` method that under the hood resolves to either `offerConstraints` or `answerConstraints` and do not even expose this separation for the user. In future major version I'd remove this separation entirely in constructor options too.

RationalCoding:
Sounds good, offer/answer can be inferred from the initiator and I don't see any use to the `setConstraints` anyways. I included it for completeness.

PR #334, opened 2018-06-29
DataChannel multiplexing

t-mullen:
Here's my implementation of DataChannel multiplexing. Opening channels and sending data works well, while preserving all of the old API.

It allows you to use the same `Peer` duplex stream as before by having the peers inherit the `DataChannel` object. But it also allows you to do multiplexing easily like this:

```javascript
var dc = peer1.createDataChannel('secondChannel')
dc // Another Duplex stream!

peer.on('datachannel', (dc) => {
  dc // Duplex stream!
})

peer.write('abc') // still a Duplex, data goes through a "default" Datachannel
```

Open Problems:
- [x] Buggy `onclose` behaviour in Chromium. Chrome intermittently doesn't  fire the event when the datachannel creator closes it but instead gets stuck on "closing".
- [x] Buggy behaviour when re-using channel names (labels). Safari throws and `InvalidStateError` when you attempt to do this. 
- [x] DataChannel instances require a reference to the original peer in order to determine if a connection is established. I didn't have much luck simply using the `onopen` event, so any help here would be appreciated.
- [x] Safari refuses to send data through a datachannel that is using the same label as a previously closed datachannel.
- [x] ~~Electron WebRTC has no support for multiple datachannels and needs to be fixed before this can be merged. https://github.com/mappum/electron-webrtc/issues/127~~ (electron-webrtc support has been dropped).
- [x] More tests are needed.

Seeing as this is another significant change, I'd appreciate feedback on the code and API from the community.


RationalCoding:
@nazar-pc As per our previous discussion about reducing bundle size, the DataChannel module can be changed to any lightweight module that has a similar constructor and exposes the 'open' event. Everything else it implements (streaming API or otherwise) will be inherited by the peer object. So a light version of the DataChannel will also make the Peer object light.

nazar-pc:
On the surface looks good, API is simple to use, will look into details later.

Are there upstream bug reports about `onopen`/`onclose`? Links would be useful if there are any.

Concerning size reduction, I think it would be a good idea to decouple basic functionality from `DataChannel` even further, so that potential `DataChannelLight` would not need to reimplement common methods. Not sure if it worth to decouple that far now though.

RationalCoding:
I'm still working on reproducing the upstream bugs with onclose simply. The onopen problem is probably an implementation flaw on my side. Definitely still some work to be done here.

I'll attempt to make a DataChannelLight implementation once these issues are worked out and we can explore further decoupling if that turns out to be difficult.

datachannel.js
```diff
@@ -0,0 +1,247 @@
...
+  self._debug('ending backpressure: bufferedAmount %d', self._channel.bufferedAmount)
+  var cb = self._cb
+  self._cb = null
+  cb(null)
+}
+
+DataChannel.prototype._onChannelOpen = function () {
+  var self = this
+  self._debug('on channel open', self.channelName)
+  self._channelReady = true
```
nazar-pc: This is only used in `Peer`, what about moving this property to `open` event handler in `Peer` constructor?

datachannel.js
```diff
@@ -0,0 +1,247 @@
...
+    self._channel.close()
+  } catch (err) {}
+
+  if (self._channel) {
+    self._channel.onmessage = null
+    self._channel.onopen = null
+    self._channel.onclose = null
+    self._channel.onerror = null
+  }
+  self._channel = null
```
nazar-pc: I'd move this under `if (self._channel)` too

datachannel.js
```diff
@@ -0,0 +1,247 @@
...
+    self._onChannelOpen()
+  }
+  self._channel.onclose = function () {
+    self._onChannelClose()
+  }
+  self._channel.onerror = function (err) {
+    self.destroy(makeError(err, 'ERR_DATA_CHANNEL'))
+  }
+
+  if (self.peer.connected) {
```
nazar-pc: Does `self._channel.readyState === 'open'` work instead of `self.peer.connected`?

nazar-pc: Implemented in https://github.com/RationalCoding/simple-peer/pull/2, addresses a third point in the first comment.

index.js
```diff
@@ -111,14 +99,25 @@ function Peer (opts) {
   // - onnegotiationneeded
 
   if (self.initiator) {
```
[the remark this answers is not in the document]
nazar-pc: Implemented in https://github.com/RationalCoding/simple-peer/pull/3

index.js
```diff
@@ -111,14 +99,25 @@ function Peer (opts) {
...
-    })
-  } else {
-    self._pc.ondatachannel = function (event) {
-      self._setupData(event)
+    var channel = self._pc.createDataChannel('default', self.channelConfig) // use label 'default' for datachannel correlation
+    self._setDataChannel(channel)
+  }
+  self._pc.ondatachannel = function (event) {
+    self._debug('ondatachannel', event.channel.label)
+    if (event.channel.label === 'default') {
```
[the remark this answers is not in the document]
nazar-pc: Implemented in https://github.com/RationalCoding/simple-peer/pull/4

nazar-pc:
Looks good to me now. How can I reproduce the first issue if it is still present?

RationalCoding:
There are already tests in this PR for those issues that were disabled for Chrome and Safari. I've turned them back on, so just running the browser tests in Chrome should be enough to reproduce.

nazar-pc:
So everything works fine in Chromium Nightly from about a week ago (a bit slow on some tests) and works perfectly smooth on latest Chromium Nightly.
Chromium stable however will often stuck on some of the tests ("data sends on seperate channels", "closing channels from creator side" and "closing channels from non-creator side" in my case).
Looks like Chromium developers have already fixed this and it is just a matter of time when this will be released. if you have Chrome Beta, would be nice to check it as well.

Shouldn't be a blocker if it is already fixed upstream.

"reusing channelNames of closed channels" will stuck in Firefox Nightly and stable, but works perfectly in both once Dev Tools are open 😕. Not sure how to debug this behavior. Dev Tools typically slows down Firefox quite a bit so probably some kind of race too. Will try to play with various timeouts.

RationalCoding:
I wasn't able to simply reproduce the issue in Chromium (it only happens around 1 in 20 times, and the obvious reproduction setup didn't work), but I did notice that the channel is always stuck in the "closing" readyState when it happens.

Added an interval to check for this condition and clean up the peer. Externally, this looks like a proper close event (since that's what should happen anyways).

nazar-pc:
Looks like everything that can be fixed/worked around on simple-peer side is ready, am I right?

RationalCoding:
Yes, everything should be good to go. I’d like to fix the electron-WebRTC issue, just haven’t had the time yet.

t-mullen:
So this has a few more unsolved browser-related issues. I've enforced unique channel names, which solved some issues, but there are more.

- [ ] Safari doesn't work correctly when data channels are created asynchronously. It announces the channel correctly, and closes correctly, but drops messages sent by the creator. The message simply never arrives on the receiving end. It's almost as if the socket never flushes, regardless of how long we wait.
- [ ] Firefox and Safari refuse to open any more channels after one is closed. I suspect they tear down everything after a single datachannel closes... 

Going to need to reproduce with native WebRTC and open bugs.

t-mullen:
I've added the `multiplex` suite of tests (minus a couple that didn't make sense for simple-peer, like chunking the multiplexed stream).

All remaining issues are browser bugs.

tinchoz49:
Hi everyone!! this is an awesome feature :+1: . Do you know when is going to be merged?

t-mullen:
Can't be merged until the browser bugs breaking some of the functionality are fixed.

t-mullen:
I've completed cross-browser testing for this.

The only bug that's left is Firefox's inability to create new channels after any are destroyed. We could support only creating DataChannels for now (no calling `dc.destroy()` before the peer is destroyed). I think this would still be very useful.

samuelmaddock:
> I think this would still be very useful.

I agree with @t-mullen here. This change moves simple-peer forward in supporting complex applications while maintaining the original public API. Would it be possible to add an `experimentalDataChannels` flag to the `Peer` options? That way it would discourage users from using it until all browser issues are resolved, but still allow for testing and more contributions to the new API.

HR:
I agree with @samuelmaddock  @t-mullen @nazar-pc this is a much-needed feature and should be added with a harmony flag so others can test it,  is it possible to do that now?

Khivar:
Hello, I took a couple of hours this morning to read your code, fork the current master branch (which is now written in ES6) and compared them. Then I adapted a little bit of code, rewrote everything in ES6 and updated the documentation. It was not really easy comparing the files manually since they didn't share a common base / were written in different ES versions so I hope I didn't forget anything. I only did some basic testing like creating another data channel, checking its label and real channel name, sending / receiving data and destroying the datachannel but so far it worked. If you want to have a look you can see the code here :

https://github.com/Khivar/simple-peer/tree/multi-datachannels

HR:
@Khivar 

I'm getting a `RangeError` when I call `peer.destroy()`

```
  simple-peer [bbed62c] destroy (error: Connection failed.) +0ms
  simple-peer [bbed62c] destroy (error: Connection failed.) +0ms
  simple-peer [bbed62c] destroy (error: Connection failed.) +0ms
  simple-peer [bbed62c] destroy (error: Connection failed.) +0ms
  simple-peer [bbed62c] destroy (error: Connection failed.) +0ms
  simple-peer [bbed62c] destroy (error: Connection failed.) +0ms
RangeError: Maximum call stack size exceeded
    at String.replace (<anonymous>)
    at debug (/Users/hr/GitHub/sd/src/main/lib/simple-peer/node_modules/debug/src/common.js:92:22)
    at Peer._debug (/Users/hr/GitHub/sd/src/main/lib/simple-peer/index.js:849:11)
    at Peer._destroy (/Users/hr/GitHub/sd/src/main/lib/simple-peer/index.js:422:10)
    at Peer.destroy (/Users/hr/GitHub/sd/src/main/lib/simple-peer/index.js:416:10)
    at /Users/hr/GitHub/sd/src/main/lib/simple-peer/index.js:425:15
    at Array.forEach (<anonymous>)
    at Peer._destroy (/Users/hr/GitHub/sd/src/main/lib/simple-peer/index.js:424:20)
    at Peer.destroy (/Users/hr/GitHub/sd/src/main/lib/simple-peer/index.js:416:10)
    at /Users/hr/GitHub/sd/src/main/lib/simple-peer/index.js:425:15
```

Khivar:
@HR 

Thanks for the feedback and sorry for the late reply. It should be fixed, I also updated the code to use the JavaScript Standard Style and synced with the upstream master branch.

samuelmaddock:
@t-mullen What are your thoughts on including this work behind an experimental flag until all browsers support it without major bugs?

t-mullen:
My concern is with interop between browsers. I'd want to run extensive testing again (you can see how many bugs came up the first time). I'd like to see interop unit tests similar to the ones in #575. I'll start work, seeing as this feature is almost 2 years overdue.

I don't understand the idea of experimental flags in an open-source project. The version of simple-peer with these features is available in this PR or in @Khivar's fork - anyone is free to use them at their own risk and they aren't officially supported.

t-mullen:
Closed in favor of #694

Issue #336, opened 2018-07-03
Remove "legacy" WebRTC API


t-mullen:
Apple has disabled the callback version of most WebRTC methods. This is what is causing the Safari 11 tests to fail.

They can be made to pass by enabling `Develop > WebRTC > Enable Legacy WebRTC API` in Travis, or by replacing the legacy API in simple-peer.

RationalCoding:
Oddly enough, Chrome displays the same "transceiver's are required" behaviour for constraints as Safari when the promise API is used (the callback version works as expected). https://bugs.webkit.org/show_bug.cgi?id=186576

fippo:
@RationalCoding how does the callback version differ from the promise one? Note that createOffer does not create state unless you call setLocalDescription

RationalCoding:
The Promise version in Chrome doesn’t update constraints on renegotiation. My guess is that it wasn’t implemented since these constraints are deprecated in favour of transceivers (which is why they aren’t supported in the Safari Promise API either). When transceivers have good support, we can switch to that and use promises everywhere.

Safari no longer supports the callback API by default, so this change is required to support it.

vblagomir:
Even with the latest commit "feross/simple-peer#90c14a8d9df1986ec1c751312e478fb70e122638" I still have the `Unhandled Promise Rejection: TypeError: Argument 1 ('options') to RTCPeerConnection.createOffer must be a dictionary` error in Safari Technology Preview 65.

vblagomir:
I have initially installed it wrongly, after reinstal the `feross/simple-peer#90c14a8d9df1986ec1c751312e478fb70e122638` works fine in the Safari Technology Preview 65. 👍

t-mullen:
This work is done.

PR #337, opened 2018-07-03
Use Promise API when possible

RationalCoding:
Implements #336.

I've changed the minimum number of methods over to the Promise API to get Safari working without the Develop flag. This will fix the tests.

Using the promise API changes the behaviour of Chrome and isn't supported in electron-webrtc, so I've included a minimal shim.

### DIFF
--- a/index.js
+++ b/index.js
@@ -86,6 +86,9 @@ function Peer (opts) {
   self._interval = null
 
   self._pc = new (self._wrtc.RTCPeerConnection)(self.config, self.constraints)
+  if (self._isChromium || opts.wrtc) {
+    shimPromiseAPI(self._wrtc.RTCPeerConnection, self._pc)
+  }
 
   // We prefer feature detection whenever possible, but sometimes that's not
   // possible for certain implementations.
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -86,6 +86,9 @@ function Peer (opts) {
   self._interval = null
 
   self._pc = new (self._wrtc.RTCPeerConnection)(self.config, self.constraints)
+  if (self._isChromium || opts.wrtc) {
```
nazar-pc: I thought Chromium supports Promise-based API for quite a while now. What is the issue there?
Also I think node-webrtc should also support Promise-based API. So maybe feature detection as in `getStats()` instead?

RationalCoding: Chromium shows the same behaviour as Safari with constraints (it ignores offerToReceiveAudio etc). Apparently this is intended behaviour as transceivers are meant to replace these deprecated options. This only happens when using the Promise API.

However, due to #95 and Chrome's lack of transceiver support, we have no other way of setting these constraints except through the callback API.

`electron-webrtc` doesn't support the Promise API. I suppose node-webrtc might be fine to use Promises.

RationalCoding: I get some false positives on other browsers with those: So I've used `self._wrtc.electronDaemon`, which should be foolproof.

nazar-pc: Doesn't look pretty, but if it works, we can live with it for a while.

test/common.js
```diff
@@ -4,7 +4,7 @@ var thunky = require('thunky')
 exports.getConfig = thunky(function (cb) {
   // Includes TURN -- needed for tests to pass on Sauce Labs
   // https://github.com/feross/simple-peer/issues/41
-  get.concat('https://instant.io/_rtcConfig', function (err, res, data) {
+  get.concat('https://instant.io/__rtcConfig__', function (err, res, data) {
```
nazar-pc: This is already in master branch

index.js
RationalCoding: Hopefully we can move to transceivers eventually and remove all callback usage. 👍

test/common.js
RationalCoding: Didn't notice that commit, thanks.

PR #345, opened 2018-07-24
'readable-stream' is useless, excessive processing

didlie:
If the feature doesn't work or does work, it makes no difference if it is removed from the options list.

totally useless library included as 'readable-stream'


nazar-pc:
It might be useless to you, but not to everyone. There are projects relying heavily on this feature.
Also code style is messed up a lot, this can't be accepted.
Once #334 is merged, it will be possible to add compact version without `readable-stream` for browsers, while still having other version for those who need it.

PR #355, opened 2018-09-10
Implement replaceTrack

CGNonofr:
Implements replaceTrack

https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSender/replaceTrack

### DIFF
--- a/index.js
+++ b/index.js
@@ -282,7 +282,7 @@ Peer.prototype.replaceTrack = async function (oldTrack, newTrack, stream) {
   if (!sender) {
     self.destroy(new Error('Cannot replace track that was never added.'))
   }
-  self._senderMap.set(newTrack, submap)
+  if (newTrack) self._senderMap.set(newTrack, submap)
 
   if (sender.replaceTrack != null) {
     await sender.replaceTrack(newTrack)


t-mullen:
Looks good, but I hit serious browser problems with replaceTrack when implementing multitrack support. They might be resolved now, but some tests for this would be fantastic.

index.js
```diff
@@ -266,6 +266,32 @@ Peer.prototype.addTrack = function (track, stream) {
...
+  if (!sender) {
+    self.destroy(new Error('Cannot replace track that was never added.'))
+  }
+  self._senderMap.set(newTrack, submap)
+
+  if (sender.replaceTrack != null) {
+    await sender.replaceTrack(newTrack)
+  } else {
+    this.removeTrack(oldTrack, stream)
+    this.addTrack(newTrack, stream)
```
t-mullen: I'm not sure about this fallback. It doesn't seem to be equivalent. ie) 'track' events will fire on some browsers, and not on others.

If people want to use browsers that don't support replaceTrack, maybe they should use removeTrack and addTrack explicitly?

t-mullen:
We don't usually push the build until an official release. Please remove this file from the PR.

CGNonofr:
Generated file removed!

CGNonofr:
I tried to add some tests, but I really don't know how to test, there is no event on the other side when a track is replaced, nor any way to know it by polling or whatever!

index.js
t-mullen: What do you think @CGNonofr?

CGNonofr: You're right, and I already changed my PR! Is throwing an error ok for you?

t-mullen: We usually use the `self.destroy` method, and it's needed here since senderMap is left in a bad state. I've made the change 👍

t-mullen:
I've tested this manually. Seem to work fine in Firefox and Chrome. `replaceTrack` support is iffy compared to removeTrack and addTrack, so I don't recommend using this unless you really need it.

I'll see if I can't PR documentation and testing soon.

PR #356, opened 2018-09-10
Peer.removeTrack does not take a sender as parameter

CGNonofr:


CGNonofr:
Indeed! I was not able to make this `removeTrack` working on firefox to chrome anyway (unlike replaceTrack, in another PR). Firefox throws this `NS_ERROR_UNEXPECTED` not only at startup, but randomly... maybe when we want to remove more than 1 track in a row.

t-mullen:
Firefox throws that error when you try calling pc.removeTrack when signalling isn't stable. Meaning the behaviour covered by this fix isn't supported by Firefox anyways 😄

PR #376, opened 2018-10-31
Feat/optional data channel

langep:
I love this project and it really makes setting up WebRTC peers easy. However, I was working with a Freeswitch server recently which for some reason cannot handle data channels (or I have not figured out how to yet). Anyway, I thought it would be a harmless addition to make the data channels optional but default using them to true. It is certainly helpful for me.


t-mullen:
We use datachannels to determine if the peers have connected.

Notice `channelReady` is set to `true` only when the datachannel opens:
https://github.com/feross/simple-peer/blob/master/index.js#L859

maybeReady will always return if `channelReady` isn't `true`.
https://github.com/feross/simple-peer/blob/master/index.js#L649

Since maybeReady will always return without a datachannel, `peer.connected` is never set to true and a large amount of necessary code is never run. You'll still get `stream` events, so your usecase is probably fine.

Just skipping the datachannel would cause a lot of this kind of unexpected behaviour. simple-peer is tightly coupled to the datachannel and it'll take some more work to change that (see related #334). If you can get the tests that do not involve the data channel to work, that'd be a start.

langep:
Fair point. I might just keep using my fork of this particular project since it sounds like simple-peer was built for the data channel.

Issue #354, opened 2018-09-06
[Question] In cases where we set trickle to false, shouldn't the generated SDP be filtered of 'a=ice-options:trickle' line? 


nikhs:
[draft-ietf-mmusic-trickle-ice-02](https://tools.ietf.org/html/draft-ietf-mmusic-trickle-ice-02#section-5.1) mentions that agents that support trickle ICE should include an  `a=ice-options:trickle` line.
I will have to verify it, but I believe most modern browsers include this line in the generated SDP in all cases  since they support trickle ICE.  

A use case where the current behavior can cause issues : Peer A and Peer B both have trickle set to false. Peer A sends an SDP that incorrectly indicates a=ice-options:trickle. Once Peer B receives it and sets it as remote description, it will immediately start generating candidates and start testing candidate pairs even though these candidates haven't been transmitted to Peer A. In Chrome and Firefox this can cause the browser to throw ICE Connection failed error after some time since candidate pair connectivity tests will exceed the timeout provided by the browser.

nazar-pc:
I think this is something we should support in simple-peer, thanks for reporting it!

t-mullen:
Just because a client does not use trickle does not mean it does not support it. For example, the initiator could disable trickle, and the non-initiator could leave it on, this is valid since both clients still support it.

This is referred to as "half-trickle":

> Half Trickle:  A trickle ICE mode of operation where the offerer
>       gathers its first generation of candidates strictly before
>       creating and sending the offer.  Once sent, that offer can be
>       processed by vanilla ICE agents and does not require support for
>       this specification.  It also allows trickle ICE capable answerers
>       to still gather candidates and perform connectivity checks in a
>       non-blocking way, thus roughly offering "half" the advantages of
>       trickle ICE.  The mechanism is mostly meant for use in cases where
>       support for trickle ICE cannot be confirmed prior to sending a
>       first offer.


The `a=ice-options:trickle` line indicates support, not usage. Even if both client indicate support, they may still choose to use "vanilla" ice where all candidates are included in the offer and answer.

>    An agent can respond to an initial offer at any point while gathering
>    candidates.  The answer can again contain any set of candidates
>    including none or **all of them**."

It's unfortunate that this causes a bug with Chrome/Firefox, but this is them not following the spec.

If you want to add this behaviour (disabling trickle by removing the support line), you can do it with an sdpTransform.

```
new Peer({
  trickle: false,
  sdpTransform: (sdp) => {
[...]

PR #378, opened 2018-11-01
Filter trickle lines when trickle is disabled.

t-mullen:
Solves #354 

Announcing trickle support and then not engaging in trickle can cause browsers to throw ` ICE Connection failed`. This PR changes the default behaviour to not announce support when trickle is disabled.

Also adds and documents `allowHalfTrickle`, which disables this filtering.

This will break applications that do SDP munging and rely on the trickle lines being present. These applications will need to set `allowHalfTrickle: true` to get the old behaviour.

### DIFF
--- a/index.js
+++ b/index.js
@@ -985,7 +985,7 @@ function shimPromiseAPI (RTCPeerConnection, pc) {
 
 // HACK: Filter trickle lines when trickle is disabled #354
 function filterTrickle (sdp) {
-  return sdp.replace(new RegExp('a=ice-options:trickle\\s\\n', 'g'), '')
+  return sdp.replace(/a=ice-options:trickle\s\n/g, '')
 }
 
 function makeError (message, code) {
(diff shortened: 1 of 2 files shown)


t-mullen:
Probable cause of Firefox test stall for disabling trickle.

index.js
```diff
@@ -980,6 +983,11 @@ function shimPromiseAPI (RTCPeerConnection, pc) {
   }
 }
 
+// HACK: Filter trickle lines when trickle is disabled #354
+function filterTrickle (sdp) {
+  return sdp.replace(new RegExp('a=ice-options:trickle\\s\\n', 'g'), '')
```
nazar-pc: Why not `/a=ice-options:trickle\s\n/g`? Seems to be easier to read.

t-mullen: Sure. I forgot that syntactic sugar with regexes. 👍

PR #383, opened 2018-11-06
Additional tests and coverage.

t-mullen:
Ran coverage over the module. Seems we have pretty good coverage, but there were a few notable things missing. This should fix that.

Only gaps left are some simple React Native logic and error handling - should be fine.

Code coverage is not wired into Travis CI yet. Any ideas?

### DIFF
--- a/test/basic.js
+++ b/test/basic.js
@@ -110,18 +110,6 @@ test('data send/receive text', function (t) {
     t.equal(peer1.initiator, true, 'peer1 is initiator')
     t.equal(peer2.initiator, false, 'peer2 is not initiator')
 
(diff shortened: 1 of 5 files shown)


test/basic.js
```diff
@@ -110,18 +110,6 @@ test('data send/receive text', function (t) {
     t.equal(peer1.initiator, true, 'peer1 is initiator')
     t.equal(peer2.initiator, false, 'peer2 is not initiator')
 
-    // TODO: re-enable after Chrome 58 is released!
```
t-mullen: These are in their own test.

PR #388, opened 2018-11-18
Add Metastream to projects list

samuelmaddock:
I'm using simple-peer in [Metastream](https://getmetastream.com/) to connect WebRTC peers using data channels. Super useful library, thanks for making it!


t-mullen:
Awesome 👍

PR #389, opened 2018-11-20
Updated simplepeer.min.js to latest code

ilDon:
I've updated simplepeer.min.js to make it easier to use the latest code as a quick fix until the latest version is published on npm


t-mullen:
Thanks, this is a good idea.

I've rebuilt it myself - maintainers aren't able to properly audit compiled builds.

PR #398, opened 2018-12-04
Check if peer is destroyed in async sendOffer, sendAnswer

### LINKED ISSUE #397: Cannot read property 'localDescription' of null
I switched to latest simple-peer from github master because I need Safari support that is missing in version from npm.

From time to time I get "Cannot read property 'localDescription' of null" exception in Chrome and Firefox:

```JS
function sendOffer () {
  var signal = self._pc.localDescription || offer
  self._debug('signal')
  self.emit('signal', {
  type: signal.type,
  sdp: signal.sdp
})
```
Looks like the callback can be executed after *Peer.destroy* call that results in self._pc = null

t-mullen:
Solves a race condition when the peer is destroyed before the createOffer promise resolves, which was throwing an uncaught error.

Fixes #397

### DIFF
--- a/index.js
+++ b/index.js
@@ -572,6 +572,7 @@ Peer.prototype._createOffer = function () {
     }
 
     function sendOffer () {
+      if (self.destroyed) return
       var signal = self._pc.localDescription || offer
       self._debug('signal')
       self.emit('signal', {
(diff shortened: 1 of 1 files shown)


jadbox:
@t-mullen no new npm or repo minimized build?

t-mullen:
@jadbox Working on getting NPM access for an active contributor. Releases should be more reliable after that.

PR #399, opened 2018-12-04
Fix tests in PRs

t-mullen:
This changes the Travis script to skip Airtap browser tests in PRs (since they always fail due to missing credentials).

Saves us the time of checking for a real build failure vs just missing credentials.


t-mullen:
Travis already does this... :/

PR #413, opened 2019-01-02
Drop electron-webrtc tests and support

t-mullen:
electorn-webrtc no longer appears to be maintained and has several open issues blocking development.

This PR removes the test suite for electron-webrtc (which is failing for a multitude of reasons). wrtc is stable and should be it's replacement.

There's no electron-webrtc dependents that have been updated within 2 years and use simple-peer:  https://www.npmjs.com/browse/depended/electron-webrtc


nazar-pc:
There are a few more places where` electron-webrtc` is mentioned, maybe remove those as well?: https://github.com/feross/simple-peer/search?q=electron-webrtc&unscoped_q=electron-webrtc

PR #418, opened 2019-01-06
Replace sender WeakMaps with normal maps

### LINKED ISSUE #407: Cannot remove track that was never added.
I am using the latest version of the library, so basically what I am trying to do is to switch video on and off in my many to many call, so my approach is:

```javascript
    // Remove current stream from peers
    this.members
      .filter(member => member.peer)
      .forEach(member => member.peer.removeStream(this.localStream));

    // Stop current stream
    this.localStream.getTracks().forEach(track => track.stop());
    this.localStream = null;

    // Adjust flags
    this.videoMode = !this.videoMode;
    if(this.videoMode) {
      this.speakerMode = true;
    }

    // Init a new stream based on the flags
    const device = this.deviceList.find(device => {
      const label = device.label.toLowerCase();
      return device.kind === 'audioinput' && this.speakerMode
        ? label.includes('speaker') || label.includes('loud')
        : label.includes('earpiece') ||
            label.includes('headset') ||
            label.includes('headphone');
    });
    this.localStream = await navigator.mediaDevices.getUserMedia({
      video: this.videoMode ? { facingMode: 'user' } : false,
      audio: device ? { deviceId: { exact: device.deviceId } } : true
    });

    // Add new stream to peers
    this.members
      .filter(member => member.peer)
      .forEach(member => member.peer.addStream(this.localStream));
```

But I am getting the following error:

```
Error: Cannot remove track that was never added.
    at r.removeTrack (simplepeer.min.js:11)
    at simplepeer.min.js:11
    at Array.forEach (<anonymous>)
    at r.removeStream (simplepeer.min.js:11)
    at call.component.ts:498
    at Array.forEach (<anonymous>)
    at CallComponent.<anonymous> (call.component.ts:498)
    at step (tslib.es6.js:97)
    at Object.next (tslib.es6.js:78)
    at tslib.es6.js:71
```

t-mullen:
WeakMaps do not prevent garbage collection and delete their key-value pair when the value is GC'd. This means stream-to-sender mappings will seem to suddenly vanish, making streams impossible to remove.

This replaces WeakMaps with normal objects, which have the same behaviour but without the GC issues.

Fixes #407


nazar-pc:
But why regular object instead of `Map`? Any specific benefits there?
`WeakMap` is specifically designed to not prevent data from being GC'ed, but `Map` doesn't have that property.

t-mullen:
No real reason. Is there a benefit to using Maps?

t-mullen:
Ah objects can't actually take objects as keys. I'll change it.

PR #420, opened 2019-01-19
Use default iceServers when none are specified in a config object.

t-mullen:
If custom `config` objects have no `iceServers` option, the default `Peer.config.iceServers` should be used.

Should reduce ICE failures caused by incomplete config objects.

### DIFF
--- a/index.js
+++ b/index.js
@@ -39,8 +39,7 @@ function Peer (opts) {
 
   self.initiator = opts.initiator || false
   self.channelConfig = opts.channelConfig || Peer.channelConfig
-  self.config = opts.config || Peer.config
-  self.config.iceServers = self.config.iceServers || Peer.config.iceServers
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -40,6 +40,7 @@ function Peer (opts) {
   self.initiator = opts.initiator || false
   self.channelConfig = opts.channelConfig || Peer.channelConfig
   self.config = opts.config || Peer.config
```
nazar-pc: What do you think about `Object.assign({}, opts.config, Peer.config)`?
Checking separate properties manually feels very hacky.

t-mullen: Sure. Shouldn't it be `Object.assign({}, Peer.config, opts.config)`?

nazar-pc: Yes, my bad

PR #438, opened 2019-02-21
handle legacy reportstats

### LINKED ISSUE #221: Can no longer connect to react-native-webrtc with simple-peer
I've been using simple-peer to establish WebRTC connections between node and a react-native app. After upgrading from 8.1.1 to 8.2.0, I'm seeing simple-peer hang on the react-native side after establishing the connection (it's stuck in connecting state, calling getStats() over and over non-stop). This means it never reaches the connected state, even thought the other side thinks it's connected.

Going through the recent commits, the breaking change was #215. From what I can tell, the cause is this:
1) Prior to #215, once we call Peer._maybeReady(), we are considered connected. After #215, we need a candidate pair that satisfies certain conditions, to be considered connected.
2) react-native-webrtc doesn't seem to return a proper candidate pair from getStats(). First, it's not even formatted correctly, it's just a giant string blob (see https://github.com/oney/react-native-webrtc/issues/396). Second, even if it's in the right format, it has a `googCandidatePair` item, but no googActiveConnection flag. This causes simple-peer to keep searching for a connected candidate pair, which will never come, hence the infinite loop.

For reference, these are the item pairs I'm getting back from react-native-webrtc's RTCPeerConnection:

```
items: Array(8)
0: {id: "googLibjingleSession_8215055649483527714", type: "googLibjingleSession", timestamp: 1514769244665.483, values: Array(1)}
1: {id: "datachannel_0", type: "datachannel", timestamp: 1514769244665.483, values: Array(4)}
2: {id: "googCertificate_CF:95:9F:42:E2:A4:DA:EA:EC:FF:77:0…D:CD:E7:5D:7D:B6:D5:A3:58:10:C6:0B:6A:76:D4:F4:37", type: "googCertificate", timestamp: 1514769244665.483, values: Array(3)}
3: {id: "Channel-audio-1", type: "googComponent", timestamp: 1514769244665.483, values: Array(2)}
4: {id: "Conn-audio-1-0", type: "googCandidatePair", timestamp: 1514769244665.483, values: Array(21)}
5: {id: "Cand-bWO5qs1f", type: "localcandidate", timestamp: 1514769244665.483, values: Array(6)}
6: {id: "Cand-u/UDIUBj", type: "remotecandidate", timestamp: 1514769244665.483, values: Array(5)}
7: {id: "bweforvideo", type: "VideoBwe", timestamp: 1514769244665.483, values: Array(7)}
length: 8
```

Just as a workaround, if I remove the check for googActiveConnection, the connection no longer gets stuck. Unfortunately, I don't understand the WebRTC spec well enough to know what's at fault here: does react-native-webrtc need to fix the output of getStats(), or should simple-peer be more robust to handle this kind of case?

dchowitz:
This PR fixes #221 which unfortunately is still a thing. I know the real fix should be made in react-native-webrtc, but since this isn't happening, and `getStats()` already handles other legacy cases, and this is a minor change, I thought I'll give it a try here.

### DIFF
--- a/index.js
+++ b/index.js
@@ -654,7 +654,7 @@ Peer.prototype.getStats = function (cb) {
     self._pc.getStats().then(function (res) {
       var reports = []
       res.forEach(function (report) {
-        reports.push(report)
+        reports.push(flattenValues(report))
       })
       cb(null, reports)
     }, function (err) { cb(err) })
@@ -664,7 +664,7 @@ Peer.prototype.getStats = function (cb) {
     self._pc.getStats(null, function (res) {
       var reports = []
       res.forEach(function (report) {
-        reports.push(report)
+        reports.push(flattenValues(report))
       })
       cb(null, reports)
     }, function (err) { cb(err) })
@@ -684,7 +684,7 @@ Peer.prototype.getStats = function (cb) {
         report.id = result.id
         report.type = result.type
         report.timestamp = result.timestamp
-        reports.push(report)
+        reports.push(flattenValues(report))
       })
       cb(null, reports)
     }, function (err) { cb(err) })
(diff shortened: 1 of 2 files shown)


t-mullen:
LGTM.

Issue #433, opened 2019-02-14
Implement transceiver support.


t-mullen:
As discussed in #430, Unified plan does not permit non-initiators to add media lines. The initiator must instead create a transceivers for each media streams it wants to receive.

We can either expose `addTransceiver` or use the out-of-band "renegotiate" signal message to call `addTransceiver` on the initiator automatically during negotiation.

An urgent subset of #351.

t-mullen:
We should also remove anything related to constraints since they don't do anything in Unified Plan.

jarretttaylor:
@t-mullen,
I saw where you had built a new version (1593ca1), so I downloaded it and tried it with my app. I wasn't sure if it were ready for testing, but I went ahead and tried it. The addStream even didn't appear to fire on the receiving side (in my app). My logs follow:

**non-initiator (adding secondary stream):**
20:46:10.059 addStream()
20:46:10.062 addTrack()
20:46:10.065 _needsNegotiation
20:46:10.072 starting batched negotiation
20:46:10.075 requesting negotiation from initiator
20:46:10.123 signal()
20:46:10.139 signalingStateChange have-remote-offer
20:46:10.161 flushing sender queue
20:46:10.177 negotiate
20:46:10.177 signalingStateChange stable
20:46:10.177 signal

**initiator:**
20:46:10.079 simplepeer.min.js:11 signal()
20:46:10.080 simplepeer.min.js:11 got request to renegotiate
20:46:10.080 simplepeer.min.js:11 _needsNegotiation
20:46:10.083 simplepeer.min.js:11 starting batched negotiation
20:46:10.083 simplepeer.min.js:11 start negotiation
20:46:10.106 simplepeer.min.js:11 signalingStateChange have-local-offer
20:46:10.107 simplepeer.min.js:11 createOffer success
20:46:10.108 simplepeer.min.js:11 signal
20:46:10.181 simplepeer.min.js:11 signal()
20:46:10.212 simplepeer.min.js:11 flushing sender queue []
20:46:10.213 simplepeer.min.js:11 negotiate
20:46:10.213 simplepeer.min.js:11 signalingStateChange stable
20:46:14.562 simplepeer.min.js:11 iceStateChange (connection: completed) (gathering: complete)
20:46:14.562 simplepeer.min.js:11 maybeReady pc true channel true

t-mullen:
Which browser?

t-mullen:
[...]

PR #440, opened 2019-02-28
Implement media transceivers.

t-mullen:
Implements basic transceiver support. #433 

- Allows initiators to add transceivers through `addTransceiver(kind, init)`.
- Allow non-initiators to add transceivers through `addTransceiver(kind, init)`, allowing them to create new m-lines and send media more media that they are receiving.
- Removes `setConstraints` and the `constraints` option, both of which are non-standard and do nothing in Unified Plan.
- Renames `offerConstraints` and `answerConstraints` options to `offerOptions` and `answerOptions` to match the standard terminology and as they are no longer "constraints" in any sense.
- Removes internal `transformConstraints` and the promise shim as they're no longer needed.
- Changes how test dummy streams are created since Chrome 70+ requires the AudioContext to be initialized by user gesture.


t-mullen:
Unfortunately it seems we need to expose `addTransceiver` since there isn't any way to tell if there are "free" transceivers to bind to. 

Basic flow of adding media from a non-initiator (when initiator has added less media tracks):
```javascript
peer.addTransceiver('video')
peer.addTrack(videoTrack, stream)
```

If the initiator is also sending media, no new transceivers need to be explicitly added:
```javascript
peer.addTrack(videoTrack, stream)
```

nazar-pc:
Can't we parse SDP to identify initial state and then count how many tracks were added on each side? If we hide `RTCPeerConnection` instance and do not expect anyone to extract it and manipulate directly, it feels like this might actually work.

t-mullen:
I tried counting tracks/transceivers on the non-initiator side and all sorts of race conditions popped up, resulting in some extra transceivers lying around.

I'm thinking it might work if the counting is done on the initiator side however (in the requestTransceiver handler).

nazar-pc:
If that works out, it would be much cleaner API

gurupras:
Is this targeted at solving scenarios where the non-initiator tries to add a stream before the initiator?

t-mullen:
@gurupras Yes.

t-mullen:
@nazar-pc I've gotten the counting streams approach to work on Firefox and Safari, but it throws an "Aw Snap" error on Chrome for >2 streams, but without any log errors... still looking into it.

t-mullen:
Alright, this is working, although I'd like to do some cross-browser testing before merging. Users shouldn't need to call `addTransceiver` at all and it's all done out-of-band.

Some other changes:
- `removeTrack` sets the sender's track to null, but there's no way to undo this. `removeTrack` is permanent and you'll have to call `track.enable = false` to disable tracks you want to re-add. Or you can add them with another stream.
- Calling `addTrack` twice throws an error instead of failing silently.
- We weren't checking that the remoteDescription was set before adding ice candidates, this isn't valid and might solve a few ice failures (these candidates were being silently discarded!).
- Used a different technique to create test streams since AudioContext needs a user event.

t-mullen:
Blocked by https://bugzilla.mozilla.org/show_bug.cgi?id=1534673 and a Safari ice candidate bug I'm still reproducing.

t-mullen:
This is ready to merge. All cross-browser issues I could find are resolved in Firefox source, Chrome 73.0.3683.86 and Safari Technology Preview Release 78.

I made some cross-browser unit tests too that I'll probably clean up and merge at a later date.

index.js
```diff
@@ -589,11 +619,32 @@ Peer.prototype._createOffer = function () {
   }).catch(function (err) { self.destroy(makeError(err, 'ERR_CREATE_OFFER')) })
 }
 
+Peer.prototype._requestMissingTransceivers = function () {
+  var self = this
+
+  ;['audio', 'video'].forEach(kind => {
```
nazar-pc: In case we already use ES6 features (like arrow functions), we might want to also use nicer loops:
```suggestion
  for (const kind of ['audio', 'video']) {
```

index.js
```diff
@@ -248,6 +245,30 @@ Peer.prototype.send = function (chunk) {
   self._channel.send(chunk)
 }
 
+/**
+ * Add a Transceiver to the connection.
+ * @param {String} kind
+ * @param {Object} init
+ */
+Peer.prototype.addTransceiver = function (kind, init) {
```
nazar-pc: So this is now called internally, can't we rename it to `_addTransceiver` as private implementation design?
Or if there is still a use case to call it explicitly, it should be documented in readme.

t-mullen: I think there's still a use-case. If you already know exactly how many transceivers you will need, there's no point in renegotiating multiple times. You could add all the transceivers you'll need right away.

Agreed it should be documented.

t-mullen:
It turns out the transceivers from `getTransceivers()` don't have an mid if they haven't been added to the connection, so no more hacky counting track lines.

An issue with removeTrack is fixed now too.

t-mullen:
node-webrtc and react-native don't support transceivers yet. Should we include this limitation in the README?

nazar-pc:
node-webrtc supports some of the features, probably not enought yet: https://github.com/node-webrtc/node-webrtc/issues/394

t-mullen:
Ok, we'll just skip requesting new transceivers for now. `addTrack` and `removeTrack` from the initiator still works, so nothing lost.

t-mullen:
There actually is transceiver support in node-webrtc `0.36`. Added some tests, but there is an issue https://github.com/node-webrtc/node-webrtc/issues/498

t-mullen:
Alright, I found a workaround, so we have full transceiver support in node-webrtc!

index.js
```diff
@@ -221,22 +218,20 @@ Peer.prototype.signal = function (data) {
...
-      new self._wrtc.RTCIceCandidate(candidate),
-      noop,
-      function (err) { self.destroy(makeError(err, 'ERR_ADD_ICE_CANDIDATE')) }
-    )
-  } catch (err) {
-    self.destroy(makeError('error adding candidate: ' + err.message, 'ERR_ADD_ICE_CANDIDATE'))
-  }
+  self._pc.addIceCandidate(new self._wrtc.RTCIceCandidate(candidate)).catch(function (err) {
+    // HACK: node-webrtc throws an incorrect error https://github.com/node-webrtc/node-webrtc/issues/498
+    if (self._pc.signalingState !== 'closed' && err.message === 'Failed to set ICE candidate; RTCPeerConnection is closed.') {
```
t-mullen: node-webrtc internally checks if `signalingState == 'closed'` when `addIceCanidate` is called, but it does this unreliably. We have to check if `signalingState` is ACTUALLY closed, and ignore any incorrect errors.

This means some candidates might be dropped, but it should be fine since this issue only occurs once per connection.

t-mullen:
I'm good to merge this if there's no further comments.

nazar-pc:
Let's do this, amazing progress!
I think one more re-build of minified version is needed.

PR #448, opened 2019-03-10
Use addTrack or addStream depending on what webRtc has implemented

### LINKED ISSUE #447: Fallback to addStream when addTrack is not supported
Hi,

Using Simple Peer together with older webrtc API such as https://github.com/oney/react-native-webrtc where addTrack is not supported. Results in stream not being created in the Peer constructor, it also results in a chrash if addStream is used directly.

On line 139 in index.js:

``` javascript
  if ('addTrack' in self._pc) {
    if (self.streams) {
      self.streams.forEach(function (stream) {
        self.addStream(stream)
      })
    }
    self._pc.ontrack = function (event) {
      self._onTrack(event)
    }
  }
```

This makes it so that the stream is not added and no feedback is given to the user of Simple Peer.

Using the peer.addStream method doesn't have the same check.

On line 275 in index.js:

``` javascript
var sender = self._pc.addTrack(track, stream)
```

Here addTrack is used without checking if addTrack exists.

I have made a workaround that enables me to send stream from android app to a chrome browser. I will make a PR shortly.

owodunni:
fixes #447

### DIFF
--- a/index.js
+++ b/index.js
@@ -136,15 +136,13 @@ function Peer (opts) {
     }
   }
 
-  if ('addTrack' in self._pc) {
-    if (self.streams) {
-      self.streams.forEach(function (stream) {
-        self.addStream(stream)
-      })
-    }
-    self._pc.ontrack = function (event) {
-      self._onTrack(event)
-    }
+  if (self.streams) {
+    self.streams.forEach(function (stream) {
+      self.addStream(stream)
+    })
+  }
+  self._pc.ontrack = function (event) {
+    self._onTrack(event)
   }
 
   if (self.initiator) {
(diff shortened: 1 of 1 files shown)


t-mullen:
I don't think we should re-add deprecated APIs. The stream API is especially unreliable. With this change won't have cross-browser compatibility, the ability to remove or replace streams, or renegotiate. 

If you want to hack support into legacy environments, you should use a polyfill to match behaviour.

Ideally, react-native-webrtc should support the modern media API: oney/react-native-webrtc#433.

nazar-pc:
I agree, either use older version or include polyfill. And ping react-native-webrtc developers to finally update their API, people have issues with it for a long time now.

owodunni:
Sounds resonable, I still don’t like that there is no feedback provided
when addTrack is not supported by webRTC in the constructor, took me a
couple of hours to find. Would it be ok to provide somekind off feedback
there?

Best Regards
Alexander Poole

t-mullen:
Latest version will throw an error when attempting to add a track or stream when `addTrack` isn't supported.

PR #454, opened 2019-04-23
Modernizing README.md

Jorropo:
Its time for ES6 and video.srcObject !
In real the es6 is very optional, I just blocked 30 min on video.srcObject and I think it can be very usefull to add this part in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,24 +66,24 @@ Let's create an html page that lets you manually connect two peers:
 var Peer = require('simple-peer')
 var p = new Peer({ initiator: location.hash === '#1', trickle: false })
 
-p.on('error', function (err) { console.log('error', err) })
+p.on('error', err => console.log('error', err))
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -143,24 +143,35 @@ Video/voice is also super simple! In this example, peer1 sends video to peer2.
...
 
-  peer2.on('stream', function (stream) {
+  peer2.on('stream', stream => {
     // got remote video stream, now let's show it in a video tag
     var video = document.querySelector('video')
-    video.src = window.URL.createObjectURL(stream)
+    
+    // This detection is optional and is just here to be retrocompatible with some olders browsers:
+    // May can be removed
+    if (video.srcObject) {
```
t-mullen: This check actually isn't correct as `srcObject` is initialized to `null`. It should be:
```javascript
if ('srcObject' in video)
```

Issue #456, opened 2019-04-25
[Safari only] "transceiverRequest" sent in a loop when the non initiator ".addStream"  


XavM:
Using the last build with "unified plan" on Safari, and adding a stream from a non initiator peer, the signal "transceiverRequest" is sent in a loop from the non initiator to the initiator, resulting in an infinity of signals being exchanged between peers (until tab is closed)

Expected behaviour : 

".addStream" > "renegotiate" > "offer" > "answer" > "transceiverRequest" > "offer" > "answer" > "on('stream')" triggered on the initiator peer 

Current behaviour on Safari : 

".addStream" > "renegotiate" > "offer" > "answer" > "transceiverRequest" > "offer" > "answer" > "transceiverRequest" > "offer" > "answer" > "transceiverRequest" >  ... etc in a loop ...  

Tested on OSX 10.14.4 + Safari 12.1  + simplepeer.min.js commit c5f8a8ded50ef354d5a0f8a3700f8817565ed032

Works fine when ".addStream"  is triggered from the initiator
Seems to work fine on Chrome/Firefox both ways 

Ex to reproduce : https://gist.github.com/XavM/78bac54fd17a1870016342df7894ed64

t-mullen:
Seems like a new bug in latest Safari. I’ll try to get a fix this weekend.

cbikar:
I noticed a similar issue on iOS 12.2. I open a session between Chrome (desktop, initiator) and Safari (iOS, non-initiator). After adding a stream, simple-peer is emitting signaling data in a loop.

t-mullen:
Can you try the latest build in this repo?

XavM:
No more endless signaling on "OSX 10.14.4 / Safari 12.1" and "iOS 12.2/Safari" with commit 3ed91bf631c4bf9c3b4bf5a85d004bb70d621707

Good job !!

Must not be related to #456, but on "iOS 12.2/Safari", the video never shows up when the stream event is fired (Opening a dedicated issue for this one : #465)

Works fine on "OSX 10.14.4 / Safari 12.1 and 12.2"

PR #463, opened 2019-05-05
Workaround for Safari 12.1 endless signaling.

t-mullen:
Workaround for #456.

Safari 12.1 wrongly shows the mid of negotiated transceivers as `null`. This is fixed in Safari 12.2 (Technology Preview). This workaround just marks the transceiver objects as negotiated.

There are some other issues in 12.1 that are fixed in 12.2. I wasn't able to fix these, so we'll just need to wait for the update.
- Safari 12.1 won't add tracks after any are removed.
- Safari 12.1 shows the wrong stream id for `ontrack` events, meaning the `stream` event will fire once for every track for streams after the first.


feross:
Sauce Labs now has Safari 12.2. Can we enable the tests that are disabled with the `TODO: Enable in Safari 12.2` comment now?

t-mullen:
Seems like these tests pass on desktop Safari now, but iOS Safari (at least on the Saucelabs emulator) still fails.

PR #466, opened 2019-05-11
Fix tests + update wrtc and devDependencies

t-mullen:
Updates `wrtc` to `0.4.1` and removes an outdated workaround that was causing a test to fail.

Aggregates Greenkeeper's devDependencies PRs (only bowser needed code changes due to a slightly more involved API).

Disables the `delayed stream on non-initiator` test on Safari/iOS. It passes on Safari 12.2 so I believe it's already fixed on their side (will take a closer look later).

Tests should pass now.

### DIFF
--- a/index.js
+++ b/index.js
@@ -226,10 +226,6 @@ Peer.prototype.signal = function (data) {
 Peer.prototype._addIceCandidate = function (candidate) {
   var self = this
   self._pc.addIceCandidate(new self._wrtc.RTCIceCandidate(candidate)).catch(function (err) {
-    // HACK: node-webrtc throws an incorrect error https://github.com/node-webrtc/node-webrtc/issues/498
-    if (self._pc.signalingState !== 'closed' && err.message === 'Failed to set ICE candidate; RTCPeerConnection is closed.') {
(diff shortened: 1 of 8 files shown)


index.js
```diff
@@ -226,10 +226,6 @@ Peer.prototype.signal = function (data) {
 Peer.prototype._addIceCandidate = function (candidate) {
   var self = this
   self._pc.addIceCandidate(new self._wrtc.RTCIceCandidate(candidate)).catch(function (err) {
-    // HACK: node-webrtc throws an incorrect error https://github.com/node-webrtc/node-webrtc/issues/498
-    if (self._pc.signalingState !== 'closed' && err.message === 'Failed to set ICE candidate; RTCPeerConnection is closed.') {
-      return self._debug('ignoring incorrect wrtc error')
```
t-mullen: wrtc ^0.3.7 has fixed this issue.

t-mullen:
Also finally fixed the saucelabs badge! 😄

t-mullen:
The `incremental multistream` test is also failing on iOS when run on Travis. I'm guessing there's some difference between my local env and saucelabs.

I'll look more into these iOS issues later.

Issue #481, opened 2019-06-20
UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'unref' of undefined


lacker:
I'm getting this error in node.js during the cleanup while the program is shutting down. I can narrow down the problem to these two lines of code:

https://github.com/feross/simple-peer/blob/master/index.js#L876

```
self._interval = setInterval(function () { self._onInterval() }, 150)
if (self._interval.unref) self._interval.unref()
```

It seems that in Node the return value of `setInterval` can sometimes be undefined, perhaps during some post-event-loop cleanup or something like that. I think the second line here should be changed to

```
if (self._interval && self._interval.unref) self._interval.unref()
```

lacker:
Inserting this code to modify `global.setInterval` works as a workaround:

```
let oldSetInterval = global.setInterval;
global.setInterval = function(...args): any {
  let answer = oldSetInterval(...args);
  if (!answer) {
    return {};
  }
  return answer;
};
```

lacker:
I submitted a one line pull request to fix this bug in https://github.com/feross/simple-peer/pull/482 - if someone could take a look that would be great!

nazar-pc:
To me PR doesn't make any sense (not claiming it is necessary wrong though, everything can happen from my experience unfortunately). `.unref()` is called synchronously after `setInterval()`, nothing in Node.js should happen between those 2 points in time that clears `self._interval`.

Can you try to post reduce as bug report in Node.js first? If this is an upstream issue, we may accept a workaround until it is fixed, but currently it looks very suspicious.

If you can provide reduced example here and exact Node.js version, that would be useful as well.

lacker:
AFAICT `self._interval` is not getting cleared, `setInterval` is just returning undefined. That is how the workaround fixes the problem.

I'll see if I can provide a reduced example. It doesn't seem to fail on every shutdown, unfortunately.

lacker:
I figured out the problem is an interaction with the `jsdom` library. In my unit testing setup using jest, `jsdom` is providing a browser-like environment, but it replaces Node's `setInterval` with a different `setInterval` function that returns undefined when it's called during the cleanup phase. (After all tests complete, but while Node still isn't exiting because it's waiting for the various timers created by simple-timer.)

see: https://github.com/jsdom/jsdom/blob/865ad590454dd345521722184bc298b32fa40810/lib/jsdom/browser/Window.js#L659

[...]

PR #482, opened 2019-06-20
handle the case where self._interval is undefined

lacker:
In some situations, Node returns undefined for `setInterval`. This just adds a check to make sure we don't access properties on undefined

### DIFF
--- a/index.js
+++ b/index.js
@@ -873,7 +873,7 @@ Peer.prototype._maybeReady = function () {
       // fallback to using setInterval to implement backpressure.
       if (typeof self._channel.bufferedAmountLowThreshold !== 'number') {
         self._interval = setInterval(function () { self._onInterval() }, 150)
-        if (self._interval.unref) self._interval.unref()
+        if (self._interval && self._interval.unref) self._interval.unref()
       }
 
       self._debug('connect')


lacker:
This PR fixes the problem that I reported in https://github.com/feross/simple-peer/issues/481

t-mullen:
I'm not against merging this if JSDom isn't being fixed. Seems like a potentially breaking change on their side.

lacker:
Yeah, I don't know how long it will take to change JSDom - this is probably a low priority fix on their end. And it seems pretty harmless to fix it here. So IMO it would be clean just to merge this fix.

nazar-pc:
I'd actually be against merging hacks like this, especially since it is possible to hack around it without changing simple-peer.

lacker:
For what it's worth, it's not that JSDom itself is so popular that simple-peer should support it per se, it's that when I was using simple-peer in conjunction with Jest unit testing, the tests were crashing frequently outside of my code, and I tracked it down to this problem. So fixing this would let simple-peer work together with Jest unit tests. That seems popular enough that it is worth compatibility IMO.

t-mullen:
Monkey patching `window.setInterval` outside of `simple-peer` is the solution until JSDom fixes this. You should do this anyways since this behaviour may break other libraries.

I know it seems easier to include this hack in simple-peer, but realize that every single JS library will need to do this as well.

https://github.com/jsdom/jsdom/issues/2617

From spec: 
> The setInterval() method must return the value returned by the timer initialization steps ...a user-agent-defined integer that is greater than zero that will identify the timeout to be set by this call in the list of active timers.

PR #495, opened 2019-07-09
Avoid multiple calls to setRemoteDescription

### LINKED ISSUE #494: Error with multiple offers per RTCPeerConnection setup
I have encountered several scenarios where `setRemoteDescription` has been called more than once. This can happen if the `sdp` message is delayed on its way from one client, to the signalhub, to the second client. This mostly happens in slow networks with a high packet drop rate (e.g. 4G).

This is what's printed when debug is turned on:
```
simple-peer [783e416] destroy (error: InvalidStateError: Failed to execute 'setRemoteDescription' on 'RTCPeerConnection': Failed to set remote offer sdp: Called in wrong state: kHaveLocalOffer) +2ms
```

corwin-of-amber:
Should fix #494.

### DIFF
--- a/index.js
+++ b/index.js
@@ -206,7 +206,7 @@ Peer.prototype.signal = function (data) {
       self._pendingCandidates.push(data.candidate)
     }
   }
-  if (data.sdp) {
+  if (data.sdp && !self._pc.remoteDescription) {
     self._pc.setRemoteDescription(new (self._wrtc.RTCSessionDescription)(data)).then(function () {
       if (self.destroyed) return
 


t-mullen:
This breaks renegotiation. There would be no way to update the remote description.

It is valid to set the remote description more than once, but it is invalid to do so when you're in the `kHaveLocalOffer` state. `setRemoteDescription` will never be called incorrectly if your signalling setup is correct. (ie not having two peers as initiator).

This is the actual issue: https://github.com/mafintosh/webrtc-swarm/issues/37

corwin-of-amber:
Thanks, I already figured out that the signaling was the problem, thanks to @nazar-pc (in #494). Sorry! I am withdrawing this PR.

PR #498, opened 2019-07-11
Call destroy on nextTick so instantiator can bind event listeners.

guanzo:
When instantiating an RTCPeerConnection throws an error, destroy the simple-peer instance on next tick so that the instantiator (in this case the `websocket-tracker.js` module inside `bittorrent-tracker`) can bind events. This way the error is caught instead of flooding the console with errors, since if an eventemitter emits an `error` event with no listeners, the error is thrown.

Current behavior inside `bittorrent-tracker/lib/client/websocket-tracker.js`:

```javascript
_createPeer (opts) {
     // ... 

    const peer = new Peer(opts) // emits 'error' event inside constructor

    peer.once('error', onError) // Too late to bind event listener
    peer.once('connect', onConnect)

    return peer

    // ... 
  }
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -92,7 +92,8 @@ function Peer (opts) {
   try {
     self._pc = new (self._wrtc.RTCPeerConnection)(self.config)
   } catch (err) {
-    self.destroy(err)
+    process.nextTick(() => self.destroy(err))
+    return
   }
 
   // We prefer feature detection whenever possible, but sometimes that's not
--- a/index.js
+++ b/index.js
@@ -92,7 +92,7 @@ function Peer (opts) {
   try {
     self._pc = new (self._wrtc.RTCPeerConnection)(self.config)
   } catch (err) {
-    process.nextTick(() => self.destroy(err))
+    setTimeout(() => self.destroy(err), 0)
     return
   }
 
--- a/test/basic.js
+++ b/test/basic.js
@@ -30,6 +30,16 @@ test('create peer without options', function (t) {
   }
 })
 
+test('can detect error when RTCPeerConstructor throws', function (t) {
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -92,7 +92,8 @@ function Peer (opts) {
   try {
     self._pc = new (self._wrtc.RTCPeerConnection)(self.config)
   } catch (err) {
-    self.destroy(err)
+    process.nextTick(() => self.destroy(err))
+    return
```
feross: Thanks for adding this early `return`. It was a bug that it was missing before.

index.js
```diff
@@ -92,7 +92,8 @@ function Peer (opts) {
   try {
     self._pc = new (self._wrtc.RTCPeerConnection)(self.config)
   } catch (err) {
-    self.destroy(err)
+    process.nextTick(() => self.destroy(err))
```
nazar-pc: Is this browserified nicely for browsers? Maybe use `setImmediate()` or `setTimeout()` instead?

feross: It is handled nicely by `browserify` but it adds 2245 bytes to the bundle size. Unfortunately, `setImmediate` is non-standard.

guanzo:
> This seems reasonable to me. Will wait to hear from the others.
> 
> A few things:
> 
> * What is `websocket-tracker`? That doesn't seem to be a valid npm package.
> * This needs a test. I wrote one and adding it to the PR.

Oops, I meant webtorrent's `bittorrent-tracker`, which contains the `WebSocketTracker` class, will edit the OP.

index.js
t-mullen: We use setTimeout for this purpose elsewhere already.

PR #507, opened 2019-07-18
Warn and continue on unsupported end candidates

### LINKED ISSUE #503: Handle ICE end of candidate
Firefox v68 shipped an update which sends an ICE end of candidates message ([bugzilla thread](https://bugzilla.mozilla.org/show_bug.cgi?id=1318167)). The offer looks like this:
```js
{candidate: {candidate: "", sdpMLineIndex: 0, sdpMid: "0"}}
```

This seems to introduce an incompatibility between the latest version of Firefox (68.0) and Chrome (75.0.3770.100). When the offer is received in Chrome, an `OperationError` is thrown.

In my own testing, this is what I'm seeing:
chrome_2019-07-13_16-56-465FQg

~~I'm using `simple-peer@9.3.0` which isn't the latest so I'll upgrade and update this thread if it fixes the problem.~~ Error persists on `simple-peer@9.4.0`.

t-mullen:
When it receives an unsupported format for an "end" ice candidate, the peer should not error.

Fixes #503


t-mullen:
Also added a test case for mDNS candidates, for sanity. They're currently ignored by browsers that don't support them.

index.js
```diff
@@ -1015,3 +1020,8 @@ function makeError (message, code) {
   err.code = code
   return err
 }
+
+function warn (message) {
+  if (console === undefined) return
+  console.warn(message)
```
nazar-pc: I'm wondering in which case `console` can be unavailable?

t-mullen: I can't think of one.

PR #510, opened 2019-07-19
Check channel._readyState on peer.connected get

### LINKED ISSUE #480: InvalidStateError: RTCDataChannel.readyState is not 'open'
I'm calling `send` on a `simple-peer` and getting this error. It's happening in Node.js, when `peer.connected` is true. I can avoid it by checking `peer._channel.readyState` myself, but shouldn't the `simple-peer` library already be doing that for me?

Here's a stack trace:

```
InvalidStateError: RTCDataChannel.readyState is not 'open'
    at RTCDataChannel.send (/home/lacker/go/src/github.com/axiom-org/axiom/ts/node_modules/wrtc/lib/index.js:48:8)
    at Peer.Object.<anonymous>.Peer.send (/home/lacker/go/src/github.com/axiom-org/axiom/ts/node_modules/simple-peer/index.js:243:17)
```

t-mullen:
Fixes #480

Since Chrome and Node.js can get stuck in the "closing" state without ever firing `channel.onclose`, we check for `channel._readyState` in an interval (a hack). It's possible an application checks `peer.connected` just before this interval fires.

PR adds a property getter to the public `peer.connected` that checks readyState.


t-mullen:
I don't really like property getters, but without a `'onclosing'` or `'onreadystatechane'` event in the datachannel, this is the best solution I could find.

feross:
Yeah, getters are not ideal but this is a reasonable workaround. :)

PR #517, opened 2019-07-29
Ignore and warn on unsupported mDNS candidate

t-mullen:
As can be seen in the tests, iOS throws an error when it gets an mDNS candidate.

The candidate should just be ignored, like unsupported empty end candidates are.

### DIFF
--- a/index.js
+++ b/index.js
@@ -237,8 +237,8 @@ Peer.prototype._addIceCandidate = function (candidate) {
   var self = this
   var iceCandidateObj = new self._wrtc.RTCIceCandidate(candidate)
   self._pc.addIceCandidate(iceCandidateObj).catch(function (err) {
-    if (!candidate.candidate) {
-      warn('Ignoring unsupported ICE end candidate.')
+    if (!candidate.candidate || candidate.candidate.indexOf('.local') !== -1) {
+      warn('Ignoring unsupported ICE candidate.')
     } else {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -237,8 +237,8 @@ Peer.prototype._addIceCandidate = function (candidate) {
   var self = this
   var iceCandidateObj = new self._wrtc.RTCIceCandidate(candidate)
   self._pc.addIceCandidate(iceCandidateObj).catch(function (err) {
-    if (!candidate.candidate) {
-      warn('Ignoring unsupported ICE end candidate.')
+    if (!candidate.candidate || candidate.candidate.indexOf('.local') !== -1) {
```
feross: Could this use `.endsWith('.local')` instead?

index.js
```diff
@@ -743,8 +743,8 @@ Peer.prototype.getStats = function (cb) {
       cb(null, reports)
     }, function (err) { cb(err) })
 
-  // Unknown browser, skip getStats() since it's anyone's guess which style of
-  // getStats() they implement.
+    // Unknown browser, skip getStats() since it's anyone's guess which style of
+    // getStats() they implement.
```
feross: Are these indents intentional?

t-mullen: Nope. I assumed it was my updated version of `standard`, but that's not the case.

PR #519, opened 2019-07-30
Fix mDNS candidate check

t-mullen:
Fixes use of `endsWith` in #517. :/

### DIFF
--- a/index.js
+++ b/index.js
@@ -237,7 +237,7 @@ Peer.prototype._addIceCandidate = function (candidate) {
   var self = this
   var iceCandidateObj = new self._wrtc.RTCIceCandidate(candidate)
   self._pc.addIceCandidate(iceCandidateObj).catch(function (err) {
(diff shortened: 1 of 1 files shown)


feross:
I just noticed an uncaught exception related to `endsWith` being called on `null`. Seems this is still not fixed.

```
(node:5801) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'endsWith' of null
    at /home/travis/build/feross/[secure]/index.js:240:57
(node:5801) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:5801) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

PR #530, opened 2019-08-06
BREAKING: Use ES class; drop `inherits` dep

feross:
Use an ES class so we can drop the `inherits` dependency. All the WebRTC environments we support have supported ES classes for a long time.

All usages of `Peer()` must be replaced with `new Peer()` now. This was always the way code examples in the readme have been written.

I know this is a lot of code to review. The only transformations I made were:

- Remove `inherits` dep.
- Use an ES class
- Convert all `self` to `this`
- Remove all lines with `var self = this`
- Convert all anonymous functions to fat arrow syntax.

For: https://github.com/brave/brave-browser/issues/5490

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,8 +1,5 @@
-module.exports = Peer
-
 var debug = require('debug')('simple-peer')
 var getBrowserRTC = require('get-browser-rtc')
-var inherits = require('inherits')
 var randombytes = require('randombytes')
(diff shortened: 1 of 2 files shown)


feross:
> I wish vanilla JS also supported had proper properties on classes

It's stage 3. Should be possible relatively soon. https://github.com/tc39/proposal-class-fields

PR #541, opened 2019-09-03
Destroy peer on connectionState 'failed'

### LINKED ISSUE #532: Chrome m76 simple-peer doesn't destroy after disconnection (nor emits 'close')
this is due to a breaking change in chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=982793

t-mullen:
The peer should be destroyed if `connectionState === 'failed'`.

Fixes #532 and any future issues that include a failed connection state without a failed ICE connection state.


tinchoz49:
Should we also check the `disconnected` and `closed` state?

t-mullen:
`disconnected` and `closed` aren't fatal. We don't currently have a use for them internally, but maybe a separate feature can expose these in the API.

feross:
Is `master` stable enough for me to do a release with this change?

longsleep:
@feross this change might make it impossible to trigger an iceRestart -  generally i think that the `_onConnectionStateChange` should also `emit` and allow the listener or via configuration to decide wether or not a connection should be destroyed on a failed connection state.

t-mullen:
@longsleep This change is related to `connectionState`, not `iceConnectionState`. It shouldn't affect ice restarts. That being said, we have always destroyed the peer when `iceConnectionState` transitions to failed - so ice restarts have never been possible with simple-peer.

If we want to allow this, we should add explicit support for ice restarts into simple-peer, not require the application to deal with new events. ie, a `restartIce: true` option, and appropriate handling of the restart.

longsleep:
Right :+1: - I think it would be nice if simple-peer could eventually make ice restarts. Either implicitly when enabled or explicitly at the application's control.

PR #543, opened 2019-09-07
Use queueMicrotask instead of setTimeout(fn, 0)

feross:
Modern browsers throttle timers severely, so setTimeout(…, 0) usually takes at least 4ms to run. Furthermore, the throttling gets even worse if the page is backgrounded.

queueMicrotask has none of these problems. But it's not available in Node 10 or earlier, and very old browsers. [`queue-microtask`](https://github.com/feross/queue-microtask) provides a no dependency, 6 line fallback for these environments.

Eventually, we can remove the package when we don't support Node 10 anymore.


feross:
9.6.0

PR #560, opened 2019-10-16
Fix for `react-native-webrtc` change to more standards-compliant beha…

### LINKED ISSUE #559: react-native-webrtc got standards-compliant and broke
Using these versions:
`simple-peer`: 9.6.0
`react-native-webrtc`: 1.75.0

This block:
https://github.com/feross/simple-peer/blob/f45d32266ed4951e1d9ac7e6b600575eaf735a0b/index.js#L705

is broken, since `getStats()` now also returns a Promise in `react-native-webrtc`. As a result, `simple-peer` never connects.

ShishKabab:
Fixes #559

### DIFF
--- a/index.js
+++ b/index.js
@@ -691,7 +691,7 @@ class Peer extends stream.Duplex {
     }
 
     // Promise-based getStats() (standard)
-    if (this._pc.getStats.length === 0) {
+    if (this._pc.getStats.length === 0 || this._isReactNativeWebrtc) {
       this._pc.getStats()
         .then(res => {
           var reports = []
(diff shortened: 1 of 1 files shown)


t-mullen:
I think it'd be better to check the react native version and use the workaround depending on that.

This PR as-is will break applications running in an older react-native version.

t-mullen:
Dropping support for older versions is also an option - if this is too difficult.

ShishKabab:
My train of though here was if existing users upgrade their dependencies, they update all of them, while new users will probably install the newest versions of everything straight away.

t-mullen:
That makes sense. LGTM.

If you or anyone using React Native interested in a future PR - getting the test suite to run in React Native would allow us to officially support it.

ShishKabab:
Yeah, we'd also like that. Currently still learning how to set this up, especially as I guess it'd have to automatically control two emulators or something...

PR #572, opened 2019-12-13
Disable remote address test on Chrome due to mDNS

t-mullen:
Chrome's mDNS implementation hides local IPs from the Javascript context. This means the remote address is unset (although it would be nice if they'd supply the mDNS address instead of leaving it empty).

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/discuss-webrtc/6stQXi72BEU/twDfpwQ4DAAJ

Local addresses can be restored by either calling `getUserMedia` before creating the peer connection, or by disabling chrome://flags > "Anonymize local IPs exposed by WebRTC."


fippo:
>  (although it would be nice if they'd supply the mDNS address instead of leaving it empty

This is done but might happen slightly later in test scenarios to avoid a particularly mean ip leak described in https://tools.ietf.org/html/draft-ietf-rtcweb-mdns-ice-candidates-04#section-3.3.1

t-mullen:
Right, I meant the mDNS name - the local address should remain hidden.

PR #576, opened 2019-12-15
Add cross-browser interoperability testing

t-mullen:
Currently, all the tests only run within a single page on the same browser. This adds tests for interoperability between different browsers. (Safari->Firefox, Firefox->Chrome, etc).

To do this, I've modified `airtap` to allow splitting a single unit test onto two different browers. (`twintap`). This works in Saucelabs and locally. 

All of these tests pass.


t-mullen:
Merged with #694

PR #577, opened 2019-12-16
Added On/Office to README.md examples

rvdleun:
Did some big refactoring on my project in the past month and was able to finally get a WebRTC implementation that I'm happy with, using simple-peer. I would like to add it to the README.md.


t-mullen:
Very cool.

PR #588, opened 2020-01-24
Fixed the function name

billyb2:
It doesn't work with new Peer, only with SimplePeer

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,8 +116,8 @@ peer-to-peer connection is established.
 ```js
 var Peer = require('simple-peer')
 
-var peer1 = new Peer({ initiator: true })
-var peer2 = new Peer()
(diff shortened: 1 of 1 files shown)


nazar-pc:
Thanks, but as readme already states:
> This exports a `SimplePeer` constructor on window. Wherever you see `Peer` in the examples below, substitute that with `SimplePeer`.

So this is done intentionally.

billyb2:
> Thanks, but as readme already states:
> 
> > This exports a `SimplePeer` constructor on window. Wherever you see `Peer` in the examples below, substitute that with `SimplePeer`.
> 
> So this is done intentionally.

Oh, my fault, I must have missed this. Thank you!

PR #601, opened 2020-03-30
Update README.md

### LINKED ISSUE #597: Video example outdated
The video sample is a bit outdated, because `navigator.getUserMedia` is deprecated.

Here is an example of a modern implementation: https://stackoverflow.com/questions/28991835/firefox-navigator-getusermedia-is-not-a-function/47958949#47958949

You might also want to mention, that the URL needs to be `https` in order to allow access to the API: https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia#Security

Following both tips this will even work on iOS and macOS  ;)

holtwick:
Modify the video stream demo to workaround deprecated getUserMedia. This version also works on iOS i.e. iPhone and iPad. Fixes #597

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,17 +148,10 @@ Video/voice is also super simple! In this example, peer1 sends video to peer2.
 var Peer = require('simple-peer')
 
 // get video/voice stream
-if (typeof navigator.mediaDevices.getUserMedia === 'undefined') {
-  navigator.getUserMedia({
-    video: true,
-    audio: true
-  }, gotMedia, () => {});
-} else {
-  navigator.mediaDevices.getUserMedia({
-    video: true,
-    audio: true
-  }).then(gotMedia).catch(() => {});
-}
+navigator.mediaDevices.getUserMedia({
+  video: true,
+  audio: true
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -148,7 +148,17 @@ Video/voice is also super simple! In this example, peer1 sends video to peer2.
 var Peer = require('simple-peer')
 
 // get video/voice stream
-navigator.getUserMedia({ video: true, audio: true }, gotMedia, () => {})
+if (typeof navigator.mediaDevices.getUserMedia === 'undefined') {
```
nazar-pc: I do not think this check is really needed, we can just switch to new API, old API is too old

holtwick: I know IE 11 is almost dead, but for those few that still need to support it, I would leave it in for another year: https://caniuse.com/#feat=mdn-api_mediadevices_getusermedia

holtwick: Fine. I'll change it. I saw the IE 11 is not supported, but it does not support WebRTC anyway :) https://caniuse.com/#feat=mdn-api_mediadevices_getusermedia

PR #632, opened 2020-04-28
fix: add buffer

hugomrdias:
This PR adds buffer as a dependency to avoid relying on bundlers automatic node globals injection.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,6 +2,7 @@ var debug = require('debug')('simple-peer')
 var getBrowserRTC = require('get-browser-rtc')
 var randombytes = require('randombytes')
 var stream = require('readable-stream')
+var { Buffer } = require('buffer')
 var queueMicrotask = require('queue-microtask') // TODO: remove when Node 10 is not supported
 
 var MAX_BUFFERED_AMOUNT = 64 * 1024
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "url": "https://github.com/feross/simple-peer/issues"
   },
   "dependencies": {
+    "buffer": "^5.6.0",
     "debug": "^4.0.1",
     "get-browser-rtc": "^1.0.0",
     "queue-microtask": "^1.1.0",


feross:
This PR should not have been merged IMO. There are several issues:

1. `require('buffer')` always uses the Node.js buffer, ignoring the `buffer` package from npm which is installed in `node_modules`. To use the npm version, you need to do `require('buffer/')`.

2. We don't actually want to use the npm version when running in Node.js since it's less efficient and compatible than the built-in Buffer.

hugomrdias:
1. Exactly but if you use bundlers without node builtins injection you get the npm package
2. This package will use node buffer when in node 

So this works exactly the same way as before plus supports bundlers without injection.

hugomrdias:
@feross any chance i can convince you to reimplement this ?
This PR is part of a larger endeavour to remove node globals and built ins from ipfs so we don't rely on bundlers to inject node stuff and prepare for webpack 5, we already worked in the maintainers from bl, readable-stream, leveldb, etc to make this a reality would be really important to have simple-peer too.

EricEisaman:
I agree.  As we prepare for [Deno](https://deno.land/) to take the stage, removing anything Node specific is a wise choice.

feross:
@hugomrdias Can you link the PRs from those other projects so we can take a look?

hugomrdias:
https://github.com/Level/level-js/pull/191
https://github.com/Level/abstract-leveldown/pull/362
https://github.com/rvagg/bl/pull/81
https://github.com/nodejs/readable-stream/pull/435 (this one is still wip but not because of buffer)

nazar-pc:
Can't we have `./buffer.js` that will re-export native Buffer in Node.js and polyfill in browsers using `browser` field in `package.json`?

hugomrdias:
The changes in the PR basically do that, but the node/browser dance is inside the buffer package.

vasco-santos:
Hey folks! Is there any update on this?

t-mullen:
@feross I disagree with the revert of this PR. 

> require('buffer') always uses the Node.js buffer, ignoring the buffer package from npm which is installed in node_modules. To use the npm version, you need to do require('buffer/').
We never used the NPM module before and this PR does not change that. The intent here is not to start using the NPM version either. There's no problem here.

> We don't actually want to use the npm version when running in Node.js since it's less efficient and compatible than the built-in Buffer.
I agree, and this PR doesn't stop using native `buffer` or intend that.

The only change here is with the browser. Currently browserify will see `require('buffer')` and automatically inject the NPM version. Not all bundlers do this and will fail due to the non-existant buffer dependency.

feross:
@t-mullen Yep, you're right. I just reverted my revert and shipped this in 9.9.0.

Issue #400, opened 2018-12-10
Not Compatible with MS Edge and Safari?


riddle28:
I am developing a chat application with simple-peer and I have tested the sample on your index page. It works fine with browsers except from MS Edge and Apple Safari because it consoles the error with error code c004e001, but chrome and firefox works fine with simple-peer. What was the reason for it? Due to unable to support them?

t-mullen:
We don't currently support MS Edge because they do not support data channels, and the WebRTC implementation is questionable in general (Microsoft prefers ORTC).

Safari should work in the latest source code and bundle.

jimmywarting:
> and the WebRTC implementation is questionable in general (Microsoft prefers ORTC).

Well, [Edge is ditching EdgeHTML and changing to use the Blink engine soon](https://hub.packtpub.com/microsoft-reportedly-ditching-edgehtml-for-chromium-in-the-windows-10-default-browser/) so 👍 
Means we will have one less annoying engine to worry about.

t-mullen:
Closing since there's no plans to support Edge until datachannels are supported.

Feel free to open another issue if you have Safari bugs.

PR #652, opened 2020-05-11
test: skip ip tests on Edge (same as Chrome)

feross:


feross:
Tests are passing again, and we're now testing Edge and Chrome on Android :)

jeremija:
@feross Is Edge supported by simple-peer? 

I had some issues with data channels recently and then found the comment below so I stopped looking for a solution, but it would be awesome if I could support Edge.

https://github.com/feross/simple-peer/issues/400#issuecomment-446373946

> Closing since there's no plans to support Edge right now.

nazar-pc:
Chromium-based Edge should just work.
Older versions do not support data channels, so can't be supported by simple-peer as is.
Also they are unfortunate legacy at this point and probably not many people care about them.

jeremija:
@nazar-pc thanks for a quick reply! I don't use Windows on a daily basis so I didn't know there was a difference. So the default Windows 10 install will have the old Edge installed? The new one needs to be downloaded and installed manually in a fresh install?

nazar-pc:
No idea, I don't use Windows either. I would guess it should auto-update itself to Chromium-based version and since some version it would ship with updated version. Do your research if it is something important to you.

jeremija:
The latest Win 10 Evaluation Edition VirtualBox image has the old one installed and a link to install the new one in Start Menu, hence my confusion:

image

feross:
I think they're still rolling out the new Edge to everyone, but it should be completed by the summer (or so I heard)

jarretttaylor:
As of this post, I am running Microsoft Windows 10 Home (v10.0.18363) and Edge reports itself as "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/81.0.4044.138 Safari/537.36"
I didn't do anything special to install it, so it had to have come as part of a regular update.

feross:
@jarretttaylor That seems to be the old Edge user agent. I know it mentions Chrome, but that's just because user agents are a mess.

The Chromium Edge uses the string "Edg" instead of "Edge".

Source: https://news.ycombinator.com/item?id=20030444

jarretttaylor:
Ahh... I see that now. My apologies for the noise.

PR #656, opened 2020-05-12
Fix replaceTrack header level

fvilers:
replaceTrack header was not of the same level than others.


feross:
Thanks!

fvilers:
It's not much ;-)

PR #660, opened 2020-05-14
Do not override code property of thrown errors

achingbrain:
When external code invoked by this module throws an error, it's passed to the `makeError` function which adds a `.code` property to the error.

I've come across a problem where the thrown error already has a `.code` property, and worse it's been marked read-only so this module then throws when it tries to overwrite the property, completely obscuring the original problem.

The change in this PR is to create a new error, set the `.code` property on it, and where the error was caused by another error being thrown, keep the wrapped error as a `.cause` property on the error we throw.

It uses the [err-code](npmjs.com/package/err-code) module to do so as a one-liner in order to keep everything concise.

This also fixes a usability  bug this module has, in that because the errors thrown by this module are all created in the `makeError` function, the stack trace of the thrown error starts there instead of where the error actually occurred, where `makeError` was invoked.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,6 +4,7 @@ var getBrowserRTC = require('get-browser-rtc')
 var randombytes = require('randombytes')
 var stream = require('readable-stream')
 var queueMicrotask = require('queue-microtask') // TODO: remove when Node 10 is not supported
+var errCode = require('err-code')
 
 var MAX_BUFFERED_AMOUNT = 64 * 1024
 var ICECOMPLETE_TIMEOUT = 5 * 1000
@@ -12,13 +13,6 @@ var CHANNEL_CLOSING_TIMEOUT = 5 * 1000
 // HACK: Filter trickle lines when trickle is disabled #354
 function filterTrickle (sdp) {
   return sdp.replace(/a=ice-options:trickle\s\n/g, '')
-}
-
-function makeError (err, code) {
-  if (typeof err === 'string') err = new Error(err)
-  if (err.error instanceof Error) err = err.error
-  err.code = code
-  return err
 }
 
 function warn (message) {
@@ -73,9 +67,9 @@ class Peer extends stream.Duplex {
 
     if (!this._wrtc) {
       if (typeof window === 'undefined') {
-        throw makeError('No WebRTC support: Specify `opts.wrtc` option in this environment', 'ERR_WEBRTC_SUPPORT')
+        throw errCode(new Error('No WebRTC support: Specify `opts.wrtc` option in this environment'), 'ERR_WEBRTC_SUPPORT')
       } else {
-        throw makeError('No WebRTC support: Not a supported browser', 'ERR_WEBRTC_SUPPORT')
+        throw errCode(new Error('No WebRTC support: Not a supported browser'), 'ERR_WEBRTC_SUPPORT')
       }
     }
 
@@ -104,7 +98,7 @@ class Peer extends stream.Duplex {
     try {
       this._pc = new (this._wrtc.RTCPeerConnection)(this.config)
     } catch (err) {
-      queueMicrotask(() => this.destroy(makeError(err, 'ERR_PC_CONSTRUCTOR')))
+      queueMicrotask(() => this.destroy(errCode(new Error('Could not create RTCPeerConnection'), 'ERR_PC_CONSTRUCTOR', { cause: err })))
       return
     }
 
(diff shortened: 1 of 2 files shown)


achingbrain:
@feross is there any chance this PR could be merged? If it's not right or you'd like something done differently, could you let me know what it is?

achingbrain:
@nazar-pc @t-mullen I don't appear to be having much luck getting any attention for this issue.  Is this PR likely to be merged?

nazar-pc:
I do not personally like either current or proposed option.

Since in most cases errors appear as a result of calling a method, I'd be more happy to see the method throwing an exception instead of having it to surface somewhere as an event instead. But this will be a significant breaking change, so I do not know.

Can you describe specific use case where current implementation conflicts with `.code` property?

achingbrain:
I see this infrequently in CI when testing the IPFS project:

```
ipfs: FAILED TESTS:
ipfs:   interface-ipfs-core tests
ipfs:     .pubsub.peers
ipfs:       ✖ "after all" hook for "should return peers for a topic - multiple peers"
ipfs:         Chrome Headless 81.0.4044.138 (Linux x86_64)
ipfs:       Error: Uncaught TypeError: Cannot assign to read only property 'code' of object '' (file:/home/travis/build/ipfs/js-ipfs/node_modules/simple-peer/index.js:26)
ipfs:  Command failed with exit code 1: karma start /home/travis/build/ipfs/js-ipfs/node_modules/aegir/src/config/karma.conf.js --files-custom --log-level error --bail true --timeout 5000
```

The test here is running in the browser and the error is a `DOMException` which has a [readonly `.code` property](https://developer.mozilla.org/en-US/docs/Web/API/DOMException/code).  I think the error is something to do with an underlying connection being closed unexpectedly, but it's hard to tell (or for the code to decide what to do) when the error gets smothered.

nazar-pc:
I'd wait to see what Thomas thinks about this, but definitely looks like we should do something about it.

t-mullen:
The issue here isn't that simple-peer is overriding the code, it's that the code is read-only. The native codes aren't very useful for debugging and I'd prefer to have the simple-peer codes when working on future issues.

Let's just construct own error object with the native error and override the code property on that:
```javascript
function makeError (err, code) {
  if (typeof err === 'string') {
     err = new Error(err)
  } else if (err.error instanceof Error) {
     err = new Error(err.error)
  } else {
    err = new Error(err)
  }
  err.code = code
  return err
}
```

No API change and no nonstandard "cause" property.

t-mullen:
I'd also prefer errors to be thrown while calling a method instead of the "error" event, but that would be a significant refactor. We should still with the current pattern and consider that later.

achingbrain:
> err = new Error(err) // <--

Is there a way of preserving the stack trace of the original error if you do it like that?  That's usually where the useful information is.

Alternatively you could just not override the code/messages and let the actual underlying error bubble up?

achingbrain:
Here's a third way, it just sets the `.code` property on the passed error but leaves the original message, type and stack intact, the `err-code` module takes care of Errors with read-only `.code` fields: https://github.com/feross/simple-peer/pull/686

achingbrain:
The CI error I pasted above goes away when I use a branch that uses #686 instead of `simple-peer@9.7.2`.  With the current `simple-peer` release I see build failures about 20% of the time, with the PR I've yet to see a failure despite having run it about 30 times in the last 24 hours.

achingbrain:
@nazar-pc @t-mullen sorry to nag but have you had any further thoughts on this?

t-mullen:
My opinion hasn't changed, I still don't see why we'd want to prioritize the native code over our own. Is there a specific error where the native code is more descriptive than what simple-peer is providing? 

The `DomException` example you give has a code of `0` - pretty useless for debugging compared to the context-aware codes simple-peer injects.

Passing the error into the Error constructor should persist the message. I have thoughts on changing how errors are thrown to preserve the stack trace, but that seems irrelevant to this issue.

t-mullen:
Fixing the attempted write to the readonly code is the main focus here. Changing which error codes are used and redesigning how errors are thrown seems out of scope for this issue.

achingbrain:
> I still don't see why we'd want to prioritize the native code over our own

Ok, great - that's what #686 does, it allows you to specify the `.code` property without crashing (as opposed this PR which does not override it).

If that's the approach preferred then this PR can be closed in favour of that one.

As you quite rightly point out, `0` is not a very useful code anyway.

> Is there a specific error where the native code is more descriptive than what simple-peer is providing?

TBH the problem that we are facing is that simple-peer takes the node process down when it attempts to override the `.code` property of `DOMException`s outside of a try block.  The debug information is nice to know, but it's the least of our problems.

> Passing the error into the Error constructor should persist the message.

The tedious thing here is the the `.message` for the `DOMException` is an empty string, which is why the error message is `Cannot assign to read only property 'code' of object ''`.  The `.stack` property on the other hand, has what actually went wrong as well as the stack trace, so if possible it would be better to preserve both `.message` and `.stack`, which is what #686 does.

PR #665, opened 2020-05-21
make addTrack and replaceTrack have same return values as spec methods

feross:
This PR makes addTrack and replaceTrack have same return values as spec methods that we're wrapping

### DIFF
--- a/index.js
+++ b/index.js
@@ -293,6 +293,7 @@ class Peer extends stream.Duplex {
       submap.set(stream, sender)
       this._senderMap.set(track, submap)
       this._needsNegotiation()
+      return sender
     } else if (sender.removed) {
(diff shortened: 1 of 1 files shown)


nazar-pc:
Well, it looks very inconsistent now with `replaceTrack` depending on conditions either throws or returns a Promise or returns `undefined` and destroys an instance with an error.
Maybe it is time to remove `sender.replaceTrack` check and just let it fail if the browser is too old?

feross:
> Maybe it is time to remove sender.replaceTrack check and just let it fail if the browser is too old?

Which browsers still don't have `sender.replaceTrack`?

nazar-pc:
Looks like none of major ones: https://caniuse.com/#search=replaceTrack

jzombie (approved):

PR #669, opened 2020-06-01
Improve default STUN URLs

feross:
- Remove the ?transport=udp parameter from the twilio STUN server URL because it's non-standard (see: https://github.com/aiortc/aiortc/issues/157)

- Use a single entry for both stun servers since they're interchangable and if one works, using the other isn't necessary


feross:
Thanks for the review @nazar-pc

Issue #670, opened 2020-06-02
Two peers sending video/audio leads to immediate renegotiation


feross:
Using `simple-peer` in a relatively simple setup. Seeing the non-initiator peer immediately request renegotiation even though everything is specified upfront in the constructor. I see two offer/answers sent from the initiator peer.

Here's my setup: Two peers, each initialized with the `stream` option, so `addTrack()` is called twice (once for video, once for audio) internally in the constructor. For the non-initiator peer, `this._isNegotiating` is already `true` because of this line in the constructor:

https://github.com/feross/simple-peer/blob/90c3dfd35ef3727b17e7b52b1c45cbc84d6ff416/index.js#L89

Therefore, for the non-initiator peer, the calls to `this._needsNegotiation()` in response to `addTrack()` cause a negotiation to be immediately queued. Once the connection is stable it triggers the re-negotiation. 

Is this intentional?

feross:
I have a draft PR that fixes the problem, but in a hacky way: https://github.com/feross/simple-peer/pull/671

Thoughts?

nazar-pc:
I agree, it seems to be a bit wrong. I'd expect renegotiation to not be triggered before the first offer was received and answer for it was generated. It is just pointless in this case.

t-mullen:
I've implemented a fix for this in #694, with tests and more debugging logs.

The causes were multiple:
- #570 changed the initial value of `_isNegotiating` to use the `channelConfig.negotiated` option (whether the datachannel is negotiated isn't relevant to the negotiation of the peer).
- `_firstStable` was set in the wrong place.
- The entire approach of dropping the non-initiator's 1st queued renegotiation request on first stable is wrong - this may be a valid request if we add a track right after creating an answer. What we actually want to do is start negotiation on both peers and immediately discard the non-initiator's 1st batched negotiation (since we know the initiator will negotiate). This was an unlikely race condition, but I've fixed it anyways.

feross:
Released as 9.8.0

PR #671, opened 2020-06-02
DRAFT: Don't immediately renegotiate for initial streams

feross:
Fixes https://github.com/feross/simple-peer/issues/670

This is a draft because I don't want to change the public API in order to fix this bug.

### DIFF
--- a/index.js
+++ b/index.js
@@ -976,6 +976,7 @@ class Peer extends stream.Duplex {
 
       this._remoteStreams.push(eventStream)
       queueMicrotask(() => {
+        this._debug('on stream')
         this.emit('stream', eventStream) // ensure all tracks have been added
       })
     })
--- a/index.js
+++ b/index.js
@@ -261,6 +261,7 @@ class Peer extends stream.Duplex {
       }
     } else {
       this.emit('signal', { // request initiator to renegotiate
+        type: 'transceiverRequest',
         transceiverRequest: { kind, init }
       })
     }
@@ -393,6 +394,7 @@ class Peer extends stream.Duplex {
       } else {
         this._debug('requesting negotiation from initiator')
         this.emit('signal', { // request initiator to renegotiate
+          type: 'renegotiate',
           renegotiate: true
         })
       }
@@ -911,6 +913,7 @@ class Peer extends stream.Duplex {
     if (this.destroyed) return
     if (event.candidate && this.trickle) {
       this.emit('signal', {
+        type: 'candidate',
         candidate: {
           candidate: event.candidate.candidate,
           sdpMLineIndex: event.candidate.sdpMLineIndex,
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -146,7 +146,7 @@ class Peer extends stream.Duplex {
 
     if (this.streams) {
       this.streams.forEach(stream => {
-        this.addStream(stream)
+        this.addStream(stream, false)
```
feross, line 149: This parameter tells `addTrack` that it should not trigger a renegotiation.

index.js
```diff
@@ -270,11 +270,11 @@ class Peer extends stream.Duplex {
    * Add a MediaStream to the connection.
    * @param {MediaStream} stream
    */
-  addStream (stream) {
+  addStream (stream, triggerNegotiate = true) {
```
feross, line 273: I don't actually want to add this to the public API, but this PR is just to demonstrate a fix to the problem

index.js
```diff
@@ -373,6 +373,9 @@ class Peer extends stream.Duplex {
   }
 
   negotiate () {
+    this._debug('negotiate')
+    this.emit('negotiate')
```
feross, line 377: Unrelated fix here: I moved this debug log and the undocumented `'negotiate'` event to fire when negotiate happens. Before, it was firing when the connection becomes stable, which seemed like a bug

index.js
```diff
@@ -882,6 +885,7 @@ class Peer extends stream.Duplex {
     if (this.destroyed) return
 
     if (this._pc.signalingState === 'stable' && !this._firstStable) {
+      this._firstStable = false
```
feross, line 888: This boolean seemed to be set in the wrong place. Before it was being set after any signalingStateChange instead of after the first stable state. I moved it inside the if statement so it only fires after first stable.

daxfrost:
Just tested this, it's working great! Looking forward to a release with this in.

t-mullen:
I've implemented a fix for this in #694, with tests and more debugging logs.

The causes were multiple:
- #570 changed the initial value of `_isNegotiating` to use the `channelConfig.negotiated` option (whether the datachannel is negotiated isn't relevant to the negotiation of the peer).
- `_firstStable` was set in the wrong place.
- The entire approach of dropping the non-initiator's 1st queued renegotiation request on first stable is wrong - this may be a valid request if we add a track right after creating an answer. What we actually want to do is start negotiation on both peers and immediately discard the non-initiator's 1st batched negotiation (since we know the initiator will negotiate). This was an unlikely race condition, but I've fixed it anyways.

index.js
t-mullen: It's meant to fire when negotiation completes, which coincides with `stable` without a queued renegotiation.

It's only used in some tests.

feross: I think it'd be better to name the event something other than `'negotiate'` since the convention is to emit debug logs and events named after methods when they're first called. Maybe it should be called `'negotiated'` then?

feross:
Superceded by #722

PR #686, opened 2020-06-23
fix: add code property to thrown errors

achingbrain:
Instead of specifying new errors with different messages and stack traces, just override the `.code` property of thrown errors.

This is an alternative approach to https://github.com/feross/simple-peer/pull/660 - instead of obscuring the original error message & stack trace, just add the `.code` property to the thrown error.


achingbrain:
Is there anything else that needs to be done here or can this be merged?  cc: @nazar-pc

dmotz:
@t-mullen this fix doesn't seem to be published on npm yet -- any idea of when the next release is going out? Frequently bitten by this bug. Thanks!

feross:
Released as 9.8.0

PR #689, opened 2020-06-24
Add TURN url example

subins2000:
There was no example of adding a TURN server in README. Added it by looking [here](https://instant.io/__rtcConfig__)


nazar-pc:
But you really need to know a bit about WebRTC to use it and there is enough information about it with examples on MDN for instance: https://developer.mozilla.org/en-US/docs/Web/API/RTCIceServer

subins2000:
That is true, closing this

Issue #19, opened 2015-05-02
support creating multiplexed channels


mafintosh:
Would be cool if simple-peer moved to the [multiplex](https://github.com/maxogden/multiplex) interface since webrtc/datachannels supports multiplexing.

I can help out with this if want since we need it in dat/friends as well :)

ngoldman:
:+1: would more easily allow for heartbeats so friends can be aware of other friends in channels

feross:
Yeah :+1: 

I like how `Peer` is a duplex stream since it makes it easier to use in place of a `net.Socket`. We could add a `createDataStream` method to get additional data channels in addition to the default one that gets set up with every simple-peer connection.

Whatever API we go with, we should make sure all the scenarios in the multiplex test suite pass.

kumavis:
another possible api, though perhaps a little asymmetrical:
create a primary data channel by default, and let the peer obj wrap this as a stream (as you do now).
have a method to then create additional streams over new data channels.

just a thought.

webglider:
What's the status of this?

feross:
@webglider Since the issue is still open, the status is what it looks like. No one has done the work yet.

RationalCoding:
Why not just pipe a multiplex stream through the simple-peer duplex?

```javascript
var plex = multiplex()
var peer = new SimplePeer()

plex.pipe(peer)
var stream1 = plex.createStream()
```

Is there any performance boost from using multiple data channels?

RationalCoding:
Is renegotiation required for creating datachannels?

webglider:
@RationalCoding Nope renegotiation is not required

RationalCoding:
Here's a proposal for an API to this:

We abstract DataChannels as their own objects wrapping an RTC DataChannel. The Peer remains a duplex stream, but it passes data to an internal DataChannel object. The peer can have a second DataChannel for renegotiation signalling if `opts.inbandSignaling` is true.

The API adds `addDataChannel(opts, cb)` that returns a new DataChannel instance in the callback.

DataChannels are duplex streams. They expose the datachannel label to correlate streams on both sides.

rvion:
some `addDataChannel` capabilities would be great !

context: I'm developping a fast-paced web multiplayer game with P2P support (where one tab acts as the server), and I really need to have 
- 1 reliable data channel for  important events `gameStart`, `kickPlayer`, etc. 
- 1 fast (non reliable) data channel for changes streaming.

---------

[...]

PR #694, opened 2020-06-30
DataChannel multiplexing

t-mullen:
Reimplementation of #334.

Remaining issues:
- [ ] Firefox won't emit datachannel events created after a channel is closed...
https://bugzilla.mozilla.org/show_bug.cgi?id=1513107

Still needs interop testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,7 +291,7 @@ If `opts` is specified, then the default options (shown below) will be overridde
 {
   initiator: false,
   channelConfig: {},
-  channelName: '<random string>',
+  channelName: 'default',
   config: { iceServers: [{ urls: 'stun:stun.l.google.com:19302' }, { urls: 'stun:global.stun.twilio.com:3478?transport=udp' }] },
   offerOptions: {},
   answerOptions: {},
@@ -362,6 +362,12 @@ Replace a `MediaStreamTrack` with another track. Must also pass the `MediaStream
 ### `peer.addTransceiver(kind, init)`
 
 Add a `RTCRtpTransceiver` to the connection. Can be used to add transceivers before adding tracks. Automatically called as neccesary by `addTrack`.
+
+### `datachannel = peer.createDataChannel(channelName, channelConfig)`
+
+Used to create additional DataChannel objects. DataChannels are instances of `stream.Duplex`.
+
+Firefox currently [does not support](https://bugzilla.mozilla.org/show_bug.cgi?id=1513107) creating new datachannels after closing any datachannel.
 
 ### `peer.destroy([err])`
 
@@ -444,6 +450,10 @@ peer.on('stream', stream => {
 ### `peer.on('track', (track, stream) => {})`
 
 Received a remote audio/video track. Streams may contain multiple tracks.
+
+### `peer.on('datachannel', function (datachannel) {})`
+
+Received an additional DataChannel. This fires after the remote peer calls `peer.createDataChannel()`.
 
 ### `peer.on('close', () => {})`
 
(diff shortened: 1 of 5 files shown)


nazar-pc:
To be completely honest, after working a with WebRTC for so long I no longer think more advanced features need to be added to simple-peer.
It is pretty easy to implement multiplexing/demultiplexing yourself if you really need it or just create multiple data channels.

stephanos:
> It is pretty easy to implement multiplexing/demultiplexing yourself if you really need it or just create multiple data channels.

Is there any prior art/example for this? I've tried it and failed. One problem for doing it over a single data channel for me was that I have a lot of binary data I'm sending but also need the _occasional_ short, "quick" message to be sent immediately. But I found that the large binary data was saturating the channel already and it took a long time for this quick message to go out. On the other hand, if multiple channels can be created with the current master, I'd love to know how people did that!

nazar-pc:
It depends heavily on your needs. You can simply create multiple native data channels with different names (and use the first one created for renegotiations only). Or since data channels are message-based, you can start each message with one byte indicating which multiplex channel is used, that was you immediately can have up to 256 possible multiplexed channels. If you want to interleave messages from different channels depending on priority that is also relatively straightforward to do.

t-mullen:
> It is pretty easy to implement multiplexing/demultiplexing yourself if you really need it
I originally felt the same way about multiplexing, but doing it in JS requires a significant amount of CPU resources even if done right. (Something I learned the hard way implementing simultaneous file transfer for FileFire). Multiple datachannels give you all that for free.

> or just create multiple data channels.
This PR just allows you to create multiple DataChannels. Calling it "multiplexing" isn't exactly accurate as it doesn't turn the Peer into a Multiplex stream.

> I no longer think more advanced features need to be added to simple-peer.
This is always something to consider. I'd hate for feature creep to make simple-peer unapproachable to a beginner. However, the core API is exactly the same and these DataChannels can be ignored completely (I expect most users to do just that). Anyone using simple-peer for an application that uses >1 stream of data or events will truly need this feature. Notable examples of this include chat clients and file transfer - the flagship usecases of WebRTC.

If complexity of API is your main concern, would you prefer if advanced features like multistreaming, multiple datachannels, etc were pulled out of the README into "advanced documentation"? I would agree with this.

The "simple" part of simple-peer, at least to me, implies the powerful features of WebRTC are simplified - not that "complex" features are cut entirely for the sake of simplicity. Renegotiation, trickle ice and multistreaming are all very complicated but simple-peer turns them into one-liner constructor arguments and method calls - which is what I love about this library.

nazar-pc:
I was misled by previous title then, I though it is "multiplexing" not multiple data channels. Makes sense then, I'll try to find time to review.

t-mullen:
> You can simply create multiple native data channels with different names (and use the first one created for renegotiations only).

This isn't currently possible with the public simple-peer API and the channels wouldn't be streams. The goal with this PR is to let you do exactly this.

t-mullen:
OK, my bad for calling it "multiplexing" - I think it was first called that in the initial feature request. It is not.

pedubreuil:
hello! any update when this could be merged ?
Thank you in advance :-)

ricardopolo:
I'm also interested in this! I'm thinking in stream video over a data channel, but continue keeping a data channel with JSON messages

HR:
@feross @nazar-pc any updates on this?

disarticulate:
showing interest here. commented on the #19

toebeann:
Is this PR dead? Would be nice to have the ability to add additional data channels to a simple-peer WebRTC connection without having to resort to snooping the codebase and meddling with private fields etc.

PR #722, opened 2020-08-08
Fix renegotiate signal event firing on first negotiation

### LINKED ISSUE #670: Two peers sending video/audio leads to immediate renegotiation

t-mullen:
Fixes #670. See also #671.

### DIFF
--- a/index.js
+++ b/index.js
@@ -47,7 +47,7 @@ class Peer extends stream.Duplex {
 
     this.initiator = opts.initiator || false
     this.channelConfig = opts.channelConfig || Peer.channelConfig
-    this.negotiated = this.channelConfig.negotiated
+    this.channelNegotiated = this.channelConfig.negotiated
     this.config = Object.assign({}, Peer.config, opts.config)
     this.offerOptions = opts.offerOptions || {}
     this.answerOptions = opts.answerOptions || {}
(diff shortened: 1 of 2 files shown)


f-r00t:
This patch fixed issues I was having with "renegiotate" triggering every time a connection was established between two peers.

abdatta:
Is this gonna be merged soon? I'm facing the same issue right now with simple peer.

yaplusplus:
It would be so nice to have this PR reviewed and merged
@feross 🙏

feross (approved):
LGTM!

feross:
Released as 9.8.0

PR #723, opened 2020-08-09
Delay destroy until next microtask.

### LINKED ISSUE #692: iceStateChange not called when onconnectionstatechange is invoked first
Browser: Chrome 83.0.4103.116

## Problem

In my application, I'd like to react to when a `failed` state change occurs to try and reconnect a user.

Looking at the chrome://webrtc-internals logs reveals the following event order:
chrome_2020-06-26_01-28-08AhIp

`iceconnectionstatechange` is logged, then `connectionstatechange`. However, these callbacks seem to be invoked in the reverse order for the simple-peer `RTCPeerConnection` instance.

chrome_2020-06-26_01-37-37MXcD

The `SimplePeer` instance is subsequently destroyed and no callback to `iceconnectionstatechange` is invoked. My intentions were to capture the `failed` state in an `iceStateChange` listener, but it's never called.

## Proposal

Since the webrtc-internal tool says `iceconnectionstatechange` is a legacy event, I'd prefer to be able to listen for `stateChange` instead of `iceStateChange`.

A new event `stateChange` should be emitted when the `connectionstatechange` callback is invoked.

Additionally, both `stateChange` and `iceStateChange` would ideally be called prior to the `SimplePeer` instance being destroyed. Maybe the call to `SimplePeer.destroy()` could be delayed by one tick?
```diff
_onConnectionStateChange () {
  if (this.destroyed) return
  if (this._pc.connectionState === 'failed') {
-    this.destroy(makeError('Connection failed.', 'ERR_CONNECTION_FAILURE'))
+    setTimeout(() => {
+      this.destroy(makeError('Connection failed.', 'ERR_CONNECTION_FAILURE'))
+    }, 0)
  }
}
```
This feels a bit dirty though so I'm unsure.

t-mullen:
Multiple concurrent events can fire from the RTCPeerConnection or RTCDataChannel. We expose some of these in the API (notable 'iceStateChange') but they will be lost if they are concurrent with an event that destroys the peer.

This PR delays destroying the peer object until the next microtask, allowing any other events to be emitted.

Fixes #692 and any related bugs.

### DIFF
--- a/index.js
+++ b/index.js
@@ -52,6 +52,7 @@ class Peer extends stream.Duplex {
     this.iceCompleteTimeout = opts.iceCompleteTimeout || ICECOMPLETE_TIMEOUT
 
     this.destroyed = false
+    this.destroying = false
     this._connected = false
 
     this.remoteAddress = undefined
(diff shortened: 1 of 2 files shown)


feross (approved):

feross:
Released as 9.9.2

PR #754, opened 2020-10-28
3 minor changes – Add 'type', Rename 'negotiate', add debug log

feross:
This PR includes 3 minor changes.

- Add 'type' property to all 'signal' messages
  - Only some signaling message have 'type' properties. The others rely on the key name that is in use. Having a 'type' in all messages improves console debug logs (since I can include the type in the output).
- Rename 'negotiate' event to 'negotiated'
  - I think this name is clearer since the event happens after negotiation.
- add a debug log when 'stream' is emitted
  - This is consistent with other events we fire where we log first.

### DIFF
--- a/index.js
+++ b/index.js
@@ -973,6 +973,7 @@ class Peer extends stream.Duplex {
 
       this._remoteStreams.push(eventStream)
       queueMicrotask(() => {
+        this._debug('on stream')
         this.emit('stream', eventStream) // ensure all tracks have been added
       })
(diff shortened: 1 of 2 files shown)


feross:
Test failures are unrelated to this PR.

feross:
Self-merging so I can get this out, but feel free to provide post-merge feedback.

feross:
Released as 9.9.3

PR #773, opened 2021-01-06
Document data type of .on('data') callback argument

jameshfisher:

[x] Documentation update
[ ] Bug fix
[ ] New feature
[ ] Other, please explain:

Current documentation is vague. The user needs to know which type to expect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -410,7 +410,7 @@ Fired when the peer connection and data channel are ready to use.
 
 Received a message from the remote peer (via the data channel).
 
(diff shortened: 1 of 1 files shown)


jzombie:
If objectMode is turned on (it is off by default; i.e. objectMode: true in the params), I believe what you pass in you will get out.

If it is not turned on, you will get a Buffer.

I added a test in my own fork which has it turned off, compared to the other tests which have it turned on: https://github.com/zenOSmosis/webrtc-peer/blob/master/test/object-mode.js

jzombie:
Merged into https://github.com/zenOSmosis/webrtc-peer

Issue #777, opened 2021-01-15
Won't connect to iOS 14 [sort of resolved]


jzombie:
**What version of this package are you using?**
9.9.3

**What operating system, Node.js, and npm version?**
iOS 14, Node.js 12.18.3, npm 6.14.6

**What happened?**
"close" event handler is called when connecting

**What did you expect to happen?**
Expected it to connect.  Works on iOS 13 and any other major browser.

**Are you willing to submit a pull request to fix this bug?**
Yes.

jzombie:
Added PR:  https://github.com/feross/simple-peer/pull/778

lol to this "standard" lib for making all sorts of code changes.

PR #778, opened 2021-01-15
Fix iOS 14 not being able to reliably connect

jzombie:
Note: The "real" change is the timeout around _onIceStateChange, but the "standard" linter did a number on a bunch of lines here

Fixed iOS 14 being able to reliably connect

[ ] Documentation update
[x ] Bug fix
[ ] New feature
[ ] Other, please explain:

Added setTimeout in _onIceStateChange

Fixes iOS 14 stability issues when connecting.

If it's not directly related to iOS 14, it seems to be related to a mix of using iOS 14 and using media streams directly.

Same implementation I have works good on iOS 13, and now a lot better on iOS 14.

This "standard" linter changed a bunch of lines around, lol.  Wasn't able to reduce the number of lines to what I had just changed, directly, by using it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -800,7 +800,7 @@ class Peer extends stream.Duplex {
           )
         )
       }
-    }, 500)
+    }, 100)
   }
 
   getStats (cb) {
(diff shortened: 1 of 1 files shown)


jzombie:
Related issue:  https://github.com/feross/simple-peer/issues/777

PR #783, opened 2021-02-11
Handle "RTCPeerConnection is gone" error on FF

Chocobozzz:
Complete stacktrace:

```
Uncaught (in promise) DOMException: RTCPeerConnection is gone (did you enter Offline mode?) PeerConnection.jsm:1153:12
    _validateIdentity resource://gre/modules/media/PeerConnection.jsm:1153
```

[ ] Documentation update
[X] Bug fix
[ ] New feature
[ ] Other, please explain:

Catch a promise rejection if the `peerIdentity` getter exists. See https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/peerIdentity for more information.

Linked to https://github.com/Chocobozzz/PeerTube/issues/3707

### DIFF
--- a/index.js
+++ b/index.js
@@ -124,6 +124,12 @@ class Peer extends stream.Duplex {
       this._onIceCandidate(event)
     }
 
+    if (typeof this._pc.peerIdentity === 'object') {
+      this._pc.peerIdentity.catch(err => {
+        this.destroy(errCode(err, 'ERR_PC_PEER_IDENTITY'))
+      })
+    }
+
     // Other spec events, unused by this implementation:
     // - onconnectionstatechange
     // - onicecandidateerror
--- a/index.js
+++ b/index.js
@@ -124,6 +124,7 @@ class Peer extends stream.Duplex {
       this._onIceCandidate(event)
     }
 
+    // HACK: Fix for odd Firefox behavior, see: https://github.com/feross/simple-peer/pull/783
     if (typeof this._pc.peerIdentity === 'object') {
       this._pc.peerIdentity.catch(err => {
         this.destroy(errCode(err, 'ERR_PC_PEER_IDENTITY'))


Chocobozzz:
Can we have a review on this? :)

feross:
This seems like a bug in Firefox. We're not using any peer identity features of the RTCPeerConnection, so it seems odd that it should be our job to listen for promise rejections on this property.

@fippo Sorry to bother you, but would you happen to know if this is a known Firefox bug?

feross (approved):
Let's merge and release this in the meantime.

feross:
Did some cursory debugging in the Firefox codebase and it appears that one of these two promises is likely throwing: https://searchfox.org/mozilla-central/source/dom/media/PeerConnection.jsm#1116-1117 It seems like this really shouldn't be exposed to the user if they're not using the peer identity features. But they don't check if the peer connection is using peer identity until *after* these promises.

fippo:
sounds odd. Maybe @jan-ivar or @nils-ohlmeier know more.

nils-ohlmeier:
Lets start with the basics here: the error is emitted here https://searchfox.org/mozilla-central/rev/200bfc652c5128011e7725fc7c201eb125d453e7/dom/media/PeerConnection.jsm#525
Which looks like some kind of life time issue to me.

The use of peerIdentity by @Chocobozzz appears to be simple feature detection sniffing to ensure that the code only gets executed on Firefox. @Chocobozzz are you using PeerIdentity at all when running into this issue?

Chocobozzz:
> @Chocobozzz are you using PeerIdentity at all when running into this issue?

We got this error using an external dependency (https://github.com/novage/p2p-media-loader) so I'm not sure. But this dependency seems to only use the [bittorrent-tracker client](https://github.com/webtorrent/bittorrent-tracker) like webtorrent does.

xinaesthete:
This may be completely spurious, but I hit an exception at this point of the code when trying to use `electron-wrtc` and followed the code comment to here:

```
18:14:43.015 Ably: EventEmitter.emit(): Unexpected listener exception: TypeError: Cannot read property 'catch' of null; stack = TypeError: Cannot read property 'catch' of null
    at new Peer (C:\code\ShemzaDigital\shemza-vite\node_modules\simple-peer\index.js:129:29)
```

Seems more like a bug with `electron-webrtc` not having the proper interface, I guess, rather than a `simple-peer` issue.

Indeed, in the `electron-wrtc` `RTCPeerConnection` code, I see:

```js
      this.peerIdentity = null // TODO: update this
```

PR #788, opened 2021-03-17
Ignore public method calls when destroying; throw when destroyed

feross:

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

When we call public methods like `signal()` but the peer is already in a `destroying` state, we might get an exception. For example, if the `RTCPeerConnection` constructor throws, we enter `destroying` but there's no way that the user can know that yet, since none of the destroy events have fired.

So, I propose that we silently disregard calls to public API methods when we're in the `destroying` state since destroy will happen on the next tick anyway. Furthermore, we should actually throw when any of these methods are called in the `destroyed` state, which were actually doing for `signal()` but not for any of the other methods. I named this error type `ERR_DESTROYED` for all methods.

This fixes this use case in `bittorrent-tracker`:

```js
const peer = new Peer()
peer.signal()
```

You can see an example stack trace in this issue: https://github.com/webtorrent/webtorrent-desktop/issues/1912

If the peer constructor throws then `this._pc` will be undefined and there's no way for us to know that. We *could* attempt to handle it by checking `this.destroying` but that actually doesn't work because there's an extra `queueMicrotask` before this.destroy() is even called. This was a mistake leftover from the introduction of `this.destroying` in this PR: https://github.com/feross/simple-peer/pull/723 We only need one `queueMicrotask` and we can just use the one in `destroy()`. This PR removes the extra call.

We don't actually want to force the user to check `this.destroying` before every call to a public API method, so we handle it internally.

### DIFF
--- a/index.js
+++ b/index.js
@@ -100,7 +100,7 @@ class Peer extends stream.Duplex {
     try {
       this._pc = new (this._wrtc.RTCPeerConnection)(this.config)
     } catch (err) {
-      queueMicrotask(() => this.destroy(errCode(err, 'ERR_PC_CONSTRUCTOR')))
+      this.destroy(errCode(err, 'ERR_PC_CONSTRUCTOR'))
       return
     }
 
@@ -180,7 +180,8 @@ class Peer extends stream.Duplex {
   }
 
   signal (data) {
-    if (this.destroyed) throw errCode(new Error('cannot signal after peer is destroyed'), 'ERR_SIGNALING')
+    if (this.destroying) return
+    if (this.destroyed) throw errCode(new Error('cannot signal after peer is destroyed'), 'ERR_DESTROYED')
     if (typeof data === 'string') {
       try {
         data = JSON.parse(data)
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -180,7 +180,8 @@ class Peer extends stream.Duplex {
   }
 
   signal (data) {
-    if (this.destroyed) throw errCode(new Error('cannot signal after peer is destroyed'), 'ERR_SIGNALING')
+    if (this.destroying) return
+    if (this.destroyed) throw errCode(new Error('cannot signal after peer is destroyed'), 'ERR_DESTROYED')
```
jhiesey, line 184: I don't like how there are now three behaviors for all public methods:
* Before `destroy` is called, calling a method has the desired effect
* Once `destroying` is set but before `destroyed` is set, calling a method does nothing
* Once `destroyed` is set it throws

Why not throw when either `destroying` or `destroyed` is set?

jhiesey (commented):
See comment

index.js
jhiesey: On second thought, I can't think of any better options without breaking backwards compatibility.

At least the `error` event gets emitted, so the client should be cleaning everything up.

jhiesey (approved):
I think this is the best solution without breaking backwards compatibility.

feross:
9.10.0

Issue #787, opened 2021-03-17
RTC peers are not garbage collected in a timely manner on Chromium-based browsers


R0quef0rt:
**What version of this package are you using?**
^9.9.3

**What operating system, Node.js, and npm version?**
This happens on both Windows 10 and Ubuntu 20.024. It's not an NPM or Node problem.

**What happened?**
This is a known bug in Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=825576

This is currently affecting my ability to use WebTorrent: https://github.com/webtorrent/webtorrent/pull/2017

RTC peers are released too slowly, causing an error that breaks both WebTorrent Desktop and the browser-only version of WebTorrent (while running in Chromium browsers). The following error will be seen in WD's "Developer > Show WebTorrent Process" menu: _Failed to construct 'RTCPeerConnection': Cannot create so many PeerConnections_

When this occurs, WebTorrent is completely unable to seed data. It is permanently broken until the process has been restarted.

This code can be used to fix the problem:

```
function forcePeerGC () {
  queueMicrotask(() => {
    let img = document.createElement('img')
    img.src = window.URL.createObjectURL(new Blob([new ArrayBuffer(5e+7)]))
    img.onerror = function () {
      window.URL.revokeObjectURL(this.src)
      img = null
    }
  })
}

setInterval(forcePeerGC, 30000)
```

**What did you expect to happen?**
Ideally, Chromium would perform proper garbage collection. But since that bug has existed for 3 years, I think it's unlikely to be fixed anytime soon. 

**Are you willing to submit a pull request to fix this bug?**
I am willing to do this, though I have tried and I am struggling to figure out how. Tests fail because the "document" object does not exist in Node, and neither does the "new Blob()" function.

DavidRadan:
Hi Guys ,
Any news regarding this BUG ?
Did you managed to solve this somehow ? 
This issue still happening in the lib i'm afraid ( still didn't applied the fix above ) .
Thanks

DavidRadan:
After applying the fix above the issue still remains !

PR #789, opened 2021-03-18
Fix RTC peer garbage collection in Chromium

ThaUnknown:

[ ] Documentation update
[X] Bug fix
[ ] New feature
[ ] Other, please explain:

Destroying a peer invokes a timeout to force GC [only on chromium]

RTC peers are released too slowly, causing an error that breaks both WebTorrent Desktop and the browser-only version of WebTorrent (while running in Chromium browsers). The following error will be seen in WD's "Developer > Show WebTorrent Process" menu: Failed to construct 'RTCPeerConnection': Cannot create so many PeerConnections, this is an issue when operating at a high amount of peers.

Honestly, this fix is so shitty I would rather have the bug, but it is needed for webtorrent to operate at high scale and it doesn't look like this issue is getting fixed anytime soon on chromium's side

This is a hack that addresses a bug in Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=825576, some cases that caused https://github.com/webtorrent/webtorrent/issues/1981 to occur, and https://github.com/feross/simple-peer/issues/787

### DIFF
--- a/README.md
+++ b/README.md
@@ -647,6 +647,9 @@ constructor. See the API docs above.
 - [simple-peer-files](https://github.com/subins2000/simple-peer-files) - A simple library to easily transfer files over WebRTC. Has a feature to resume file transfer after uploader interruption.
 - [WebDrop.Space](https://WebDrop.Space) - Share files and messages across devices. Cross-platform, no installation alternative to AirDrop, Xender. [Source Code](https://github.com/subins2000/WebDrop)
 - [Speakrandom](https://speakrandom.com) - Voice-chat social network using simple-peer to create audio conferences!
+- [Deskreen](https://deskreen.com) - A desktop app that helps you to turn any device into a secondary screen for your computer. It uses simple-peer for sharing entire computer screen to any device with a web browser.
+
+
 
 - *Your app here! - send a PR!*
 


index.js
```diff
@@ -429,7 +436,11 @@ class Peer extends stream.Duplex {
 
       if (!this._readableState.ended) this.push(null)
       if (!this._writableState.finished) this.end()
-
+      if (typeof window === 'object' && typeof window.chrome === 'object') {
+        if (!forcedGCTimeout) {
+          forcedGCTimeout = setTimeout(forcePeerGC, 5000)
```
DiegoRBaquero, line 441: why 5 seconds? to allow multiple destroy's to be cleaned up at once?

We should also unref the timer

ThaUnknown: honestly, this is a really stupid fix, for a really stupid issue, there are probably way better ways of doing this than what I showed here, which doesn't always work, but yeah, that timeout is to allow multiple peers to be destroyed at once.

DiegoRBaquero: This is the suggested solution for the mean time, and I believe this is the correct place to handle such "hack".

To not call this so often, I'd even go with 10 seconds.

My only nit requirement: unref the timer upon creation, shouldn't not block process close if it's the last thing standing.

ThaUnknown: unrefed the timeout, but idk about the 10s timeout, 5s is already a lot because in theory it should happen every time a peer is destroyed, at very high amt of peers which are rapidly connecting to the client, 10s might be too long, the client will exceed the wrtc conn limit and essentially kill all future wrtc connections

DiegoRBaquero (approved):

PR #796, opened 2021-04-13
Better DataChannel error handling

feross:
Don't emit an RTCErrorEvent since that's not an Error. Instead, emit the actual RTCError, if it exists. Otherwise, make a new Error and emit that.

[ ] Documentation update
[x] Bug fix
[ ] New feature
[ ] Other, please explain:

Don't emit an RTCErrorEvent since that's not an Error. Instead, emit the actual RTCError, if it exists. Otherwise, make a new Error and emit that.

None

Does this work in all browsers?

### DIFF
--- a/index.js
+++ b/index.js
@@ -528,7 +528,10 @@ class Peer extends stream.Duplex {
     this._channel.onclose = () => {
       this._onChannelClose()
     }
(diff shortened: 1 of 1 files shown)


jhiesey (approved):
LGTM; good catch

